PR #6, opened 2020-01-14
Use NPM and Webpack to build webroot

ewized:
This breaks out the JavaScript code into separate files and bundles the project into the preexisting zip.

I tried to maintain the same file structure that currently exists that can be changed since the files are being bundled together.

With that change the javascript uses es6 syntax where I briefly made the changes where I saw fit, but there is much more there can be done to get the performance tweaks of using the new syntax.

The preprocessor us typescript but did not convert any of the files to use such only using it to gain the newer syntax features of javascript.

The css has also been broken out into scss modules.

Though due note since this is a very large commit I may have missed a few things. 

I also came across formatting issues, I noticed the code has a mixed of tabs and spaces. I tried to match what was existing.

There may be more to go though but this is a thorough start


TBlueF (commented):
First of all, thank you for the contribution!

As you have noticed and addressed with this changes, i am more into Java than Web-Development and currently don't have enough js/node/web knowledge to set everything up perfectly and use the latest best-practices ^^

This looks really promising!
Though, i have some questions, so i can test and work with this setup:

Currently the build-requirements were only `java 8` and the build process was only `clone` and `./gradlew shadowJar`.
With your changes i am getting a [build-error](https://gradle.com/s/3cmq7p2d57474), i am obviously missing some requirements.
1. What are those requirements and how do i install them correctly?
1. Can/should they be added to the gradle-build-process to keep building as easy as possible?

ewized:
I see from your logs your on windows I haven't tested it on windows and will test it on windows later, but the goal was nothing should of changed from `./gradlew shadowJar`

ewized:
Are you using OpenJDK or Oracle trying to replicate your environment

TBlueF:
I am using oracle, 
~~i am surprised by how this is affecting the build process here? xD~~

ewized:
It turns out to be gradle 5, the plugin `com.moowork.node` that handles the node from gradle is just not updated yet, so I had to backport to gradle 4.9 and the shadow plugin. Also was missing a few things in `package.json` and the gradle task.

Also added that `shadowJar` depends on `assemble` task so you can just use `./gradlew build`

ewized:
That will build everything that way it was, but if you want to do development on the webroot without using gradle, you need nodejs. Then inside the folder BlueMapCore you run `npm run build` to create the web bundle, and its output will be in `BlueMapCore/build/generate/webroot/`. One of the nice things is the dev server with hotswaping which you run with `npm run start` and it will host a local webserver by the config localhost:8080 and reload the webpage everything there is a change to the javascript or css files. Since its running its own web server and bluemap looks for a folder `data/` at the root, a generated data folder should exist in `BlueMapCore/build/generate/world/` inside that folder will be the `data/` folder.

TBlueF:
Now, with the downgrade of gradle and the shadow plugin, if i compile with the sponge-plugin (uncomment [this](https://github.com/ewized/BlueMap/blob/feature/webpack/build.gradle#L31)) it completely includes the whole SpongeAPI and its dependencies in the jar.. which is bad ^^

But:
You stated that the `com.moowork.node` plugin is not compatible with gradle 5? 
Because i tried to update gradle and the shadow plugin again. And with your change to depend on the `npmInstall` task, the build runs without any issues for me :)

If you see no issue in going back to gradle 5 and the newer shadow-plugin version i will merge this.

BlueMapSponge/build.gradle
```diff
@@ -1,5 +1,5 @@
 dependencies {
-    shadow "org.spongepowered:spongeapi:7.1.0-SNAPSHOT"
+    compile "org.spongepowered:spongeapi:7.1.0-SNAPSHOT"
```
TBlueF, line 2: This results in including the whole SpongeAPI and its dependencies in the jar

ewized: Yeah, if gradle 5 works for you. You can undo some of the downgrades on my pr branch, and merge. Like I stated earlier I did not get the issue on my mac so it may just be a fluke.

PR #19, opened 2020-01-18
Limit raycast objects to the current tile

Katrix:
Saw a note about the optimization, and thought I'd go for it, as it's a big bottleneck in one of our maps. As a quick fix currently I just use the low res scene for the height scene, but this is probably better.


TBlueF:
Awesome! Looks good, thx :)

PR #62, opened 2020-07-29
Use binary data files

### LINKED ISSUE #60: [Feature Request] Use binary data files instead of json
Currently json is used to store the map data, and it's bandwidth consuming (ie. sometimes use a 20bytes String to represent a Float32). This is especially obvious when load overview (lowres data), since hires data is more "regular" and can be gzipped to a smaller size.

Also it may improve performance since there's no transform from binary(arraybuffer) to data(float32array), instead of String to float32.

djytw:
A demo implementation of  #60 

No improvements in bandwidth, not sure about performance.

### DIFF
--- a/BlueMapCore/src/main/java/de/bluecolored/bluemap/core/render/lowres/LowresModelManager.java
+++ b/BlueMapCore/src/main/java/de/bluecolored/bluemap/core/render/lowres/LowresModelManager.java
@@ -24,11 +24,11 @@
  */
 package de.bluecolored.bluemap.core.render.lowres;
 
+import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
@@ -37,8 +37,6 @@ import java.util.Map.Entry;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.zip.GZIPInputStream;
-
-import org.apache.commons.io.IOUtils;
 
 import com.flowpowered.math.vector.Vector2i;
 import com.flowpowered.math.vector.Vector3d;
@@ -172,7 +170,7 @@ public class LowresModelManager {
 	 * Returns the file for a tile
 	 */
 	public File getFile(Vector2i tile, boolean useGzip){
-		return FileUtils.coordsToFile(fileRoot, tile, "json" + (useGzip ? ".gz" : ""));
+		return FileUtils.coordsToFile(fileRoot, tile, "bin" + (useGzip ? ".gz" : ""));
 	}
 	
 	private LowresModel getModel(UUID world, Vector2i tile) {
@@ -189,10 +187,20 @@ public class LowresModelManager {
 						try (FileInputStream fis = new FileInputStream(modelFile)) {
 							InputStream is = fis;
 							if (useGzip) is = new GZIPInputStream(is);
+							byte[] buffer = new byte[1024];
 							
-							String json = IOUtils.toString(is, StandardCharsets.UTF_8);	
+							ByteArrayOutputStream out = new ByteArrayOutputStream();
+
+							int len;
+							while ((len = is.read(buffer)) > 0) {
+								out.write(buffer, 0, len);
+							}
+
+							is.close();
+							out.close();
+							byte[] data = out.toByteArray();
 							
-							model = new CachedModel(world, tile, BufferGeometry.fromJson(json));
+							model = new CachedModel(world, tile, BufferGeometry.fromBinary(data));
 						} catch (IllegalArgumentException | IOException ex){
 							Logger.global.logError("Failed to load lowres model: " + modelFile, ex);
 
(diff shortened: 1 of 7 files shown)


TBlueF:
Looks nice!
But how are there no improvements in bandwidth?

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/threejs/BufferGeometry.java
```diff
@@ -127,6 +131,109 @@ public String toJson() {
...
+		header[0] = current;
+		header[1] = encodedGroups.length;
+		current +=  encodedGroups.length;
+		int j = 2;
+		for (Entry<String, BufferAttribute> entry : this.attributes.entrySet()) {
+			header[j ++] = current;
+			byte[] data = entry.getValue().toBinary();
+			header[j ++] = data.length;
+			current += data.length;
+			int type = AttributeType.valueOf(entry.getKey()).ordinal();
```
TBlueF, line 160: This might throw an `IllegalArgumentException` if there is an attribute not listed in `AttributeType`

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/threejs/BufferGeometry.java
```diff
@@ -127,6 +131,109 @@ public String toJson() {
...
+			header[j ++] = type;
+			encodedAttrs[type] = data;
+			header[j ++] = entry.getValue().isNormalized()?1:0;
+			header[j ++] = entry.getValue().getItemSize();
+		}
```
TBlueF, line 165: Is there something speaking against moving the `isNormalized` and `itemSize` properties into the `toBinary()` method of the buffer-attribute? To me this would seem more logical.

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/render/lowres/LowresModelManager.java
```diff
@@ -189,10 +187,20 @@ private LowresModel getModel(UUID world, Vector2i tile) {
...
+							ByteArrayOutputStream out = new ByteArrayOutputStream();
+
+							int len;
+							while ((len = is.read(buffer)) > 0) {
+								out.write(buffer, 0, len);
+							}
+
+							is.close();
+							out.close();
+							byte[] data = out.toByteArray();
```
TBlueF, line 201: Since we are depending on apache.commons we could just use `byte[] data = IOUtils.toByteArray(is);` here

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/threejs/BufferAttribute.java
```diff
@@ -86,6 +89,14 @@ public void writeJson(JsonWriter json) throws IOException {
 		json.endObject();
 	}
 
+	public byte[] toBinary() {
+		ByteBuffer buffer = ByteBuffer.allocate(4 * values.length);
+		buffer.order(ByteOrder.LITTLE_ENDIAN);
+		FloatBuffer floatBuffer = buffer.asFloatBuffer();
+		floatBuffer.put(this.values);
+		return buffer.array();
+	}
```
TBlueF, line 98: I think `itemSize` and `normalized` should be included here?

BlueMapCore/src/main/webroot/js/libs/BlueMap.js
```diff
@@ -515,17 +515,17 @@ export default class BlueMap {
 
 				resolve(object);
 			}, () => {
-			}, reject);
+			}, (msg)=>{console.log(msg)});
```
TBlueF, line 518: Leftover debug? Or is there a reason for this change? :D

TBlueF (changes requested):
One thing i'd like to change is the use of an enum with fixed attribute-names:
Attributes might always be added or removed if a new feature requires them. Storing the names of the attributes in the file will only have a really small impact on the file-size but would be a big improvement in scalability.
What do you think?

TBlueF:
Oh! Also: Since those changes are for all minecraft-versions, we need to pull those changes over to the `mc/1.13` branch because i can only merge them from there.
I really should document this somewhere: The `mc/1.13` branch is the root branch from where all changes are merged into master (1.16), mc/1.12 and mc/1.15.

BlueMapCore/src/main/webroot/js/libs/BlueMap.js
djytw: My fault

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/threejs/BufferGeometry.java
djytw: Changing to string attribute name would solve this

djytw:
Sorry for not responding a while, I tested some maps and found it's not a good idea to use binary.
- most of maps will need 150% of space for storage/bandwidth when use binary, especially the hi-res maps. Not sure why, maybe json is easier gzipped to a smaller size. 
- For the performance: The bottleneck is rendering, parse a 100kb json not really takes much time.  
So I'm closing this pr

TBlueF:
Also, json is more flexible, so i agree with that decision if there are no space/performance improvements.
Still, thanks a lot for all the effort :)

Issue #98, opened 2020-09-29
Log saving for CLI


ChunkyPanda03:
Could you add the ability to have the CLI version of Blue Map save it's logs as well as being able to see real time connections to the server.

maybe like how python (python -m http.server) shows when someone connects to it

PR #100, opened 2020-10-01
Add python style request logging

rye761:
Addresses part of #98 by adding Python style request logging to the server.

BlueMap makes many requests so the log is often very long - this may not be an optimal way to see connections to the server.

This is an initial pull request to get feedback - I am open to changing what is logged or provide an option because I think this is a good idea but currently this basically spams the terminal window any time someone connects.

logging

### DIFF
--- a/BlueMapCore/src/main/java/de/bluecolored/bluemap/core/webserver/HttpConnection.java
+++ b/BlueMapCore/src/main/java/de/bluecolored/bluemap/core/webserver/HttpConnection.java
@@ -34,6 +34,9 @@ import java.net.SocketTimeoutException;
 import java.util.concurrent.TimeUnit;
 
 import de.bluecolored.bluemap.core.logger.Logger;
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.util.Date;
 
 public class HttpConnection implements Runnable {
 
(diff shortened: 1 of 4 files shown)


TBlueF:
So, i also like the idea of adding a way to display / log the connections to the webserver.
However, I'd suggest to make this an optional command-line flag (maybe `-v` / `--verbose`?), that - if being used with the `-w` option to start the webserver - enables the extra logging.
The way it is currently implemented would not only output this on the cli-version .. but also on plugins that use the internal webserver. Which is bad :D

rye761:
Alright, added it as a command line option. I use -b for the shorthand because -v was already taken for --mc-version.

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/webserver/HttpConnection.java
```diff
@@ -66,6 +72,20 @@ public void run() {
...
+                                    Logger.global.logInfo(
+                                            connection.getInetAddress().toString() +
+                                            " [ " +
+                                            dateFormat.format(date) +
+                                            " ] \"" +
+                                            request.getMethod() +
+                                            " " + request.getPath() +
+                                            " " + request.getVersion() +
+                                            "\" " +
+                                            response.getStatusCode().toString());
```
TBlueF: I think this should be in a new method `if (verbose) log(request, response)` to not clutter the main connection-loop :)

TBlueF (changes requested):
One last thing before i merge this:
Maybe you could fix the indentation to be tabs to match the rest of the project. (You used spaces)

Rest is looking good :)

*(Sorry for the late response)*

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/webserver/HttpConnection.java
rye761: No problem. I'll get this done soon.

TBlueF:
Okay, i merged this in the bleeding branch and did the rest of the formatting.
So this will be included in the next snapshot :)
*(bleeding will be merged in master as soon as i decide to release a new stable version ^^)*

rye761:
Awesome, thanks! For future pull requests I'll try and make sure the formatting is better and request to merge into bleeding instead. You may wish to add those points to your contributing details in README or make a CONTRIBUTING.md. Also, you may want to add the hacktoberfest topic to your repo. They changed the rules this year and now it's opt in. It wasn't at the start though, so this PR counted for me but to have any future PRs count the repo needs to have the tag. I'm doing the challenge and would love to make another PR sometime this month that counts toward the challenge.

PR #125, opened 2020-12-26
Clearer wording on the file download message

pop4959:
Saw a mistake with "of" instead of "for" but then thought might as well make the message read a bit more clearly


pop4959:
not sure if master is the correct branch to merge to so make sure to change that if not

PR #228, opened 2021-07-31
Migrating bug_report to github's form syntax

combe15:


TBlueF:
While i like the forms generally, I am not sure if i want to have it for bug-issues.
The template should encourage people to invest some time investigating, making sure their Bug is actually a Bug and not an error on their side. Filling out a Form makes this process way to easy and fast..
I'll think about this a bit more, leaving this PR open for now :)

TBlueF:
Won't use this for now..

PR #275, opened 2021-12-15
Add support for fabric-permissions-api

NikitaCartes:
Api provided by https://github.com/lucko/fabric-permissions-api and supported by LuckPerms


TBlueF:
Awesome!
Does this add a hard plugin-dependency to bluemap-fabric?
Or is the API shaded into the jar? :)

NikitaCartes:
It's need to be included otherwise it will not work without LuckPerms, forgot about that.
What's the best way to do it:

add `modImplementation(include("me.lucko:fabric-permissions-api:0.1-SNAPSHOT"))` in `dependencies`
or
add `include "me.lucko:fabric-permissions-api:0.1-SNAPSHOT"` in `shadowInclude`

Tested in 1.16.5, 1.17.1 and 1.18.1 with and without LuckPerms and all works fine.

TBlueF:
> What's the best way to do it

i have no idea, tbh .. whichever works best

PR #307, opened 2022-05-31
Forge 1.18.2 Support

Oscariremma:
Add Gradle build files for Forge 1.18.2

### DIFF
--- a/implementations/forge-1.18.2/build.gradle
+++ b/implementations/forge-1.18.2/build.gradle
@@ -1,0 +1,73 @@
+buildscript {
+    repositories {
+		maven { url = 'https://maven.minecraftforge.net' }
+        mavenCentral()
+    }
+    dependencies {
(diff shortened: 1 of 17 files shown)


TBlueF:
What is this for?
`BlueMap-1.7.2-forge-1.18.1` does still work with Forge 1.18.2, doesn't it?

Oscariremma:
Hi, my bad!

It does indeed appear as if  `BlueMap-1.7.2-forge-1.18.1` works for 1.18.2.

The reason I tried to compile it for 1.18.2 was that I previously had problems with 1.18.1 mods not working in 1.18.2. So when I saw Bluemap was for 1.18.1 on CurseForge my first thought was to se if I could get it to compile for 1.18.2 which was successful. And after that I thought that I might as well make a pull request with the change. What I probably should have done first is just try it but I think my curiosity got the better of me :).

So I will just close this pull request, but thanks for a really awesome and impressive mod!

TBlueF:
Well, thanks for the effort anyways! :)

PR #320, opened 2022-08-14
Docker container for CLI version

Chicken:
Idk why I did this tbh- I was fine with my selfbuilt one... I wasted so much time fighting against ghactions I now hate my life. It should work... probably...
It works by publishing the cli jar built for each commit as a docker container. It tags the images with the name of the branch and also as `latest` for master.


TBlueF:
Okay, sounds fancy!
You will now need to explain to an absolute docker-noob what for example you can do with such a docker image :)

TBlueF:
Okay, sounds fancy!
You will now need to explain to an absolute docker-noob what for example you can do with such a docker image :)

Chicken:
Docker is indeed one of those fancy tools some sysadmins like myself like to use.
Essentially the application comes prepackaged in a container with all its dependencies.
This means that one can run the application without downloading the jar, or even java!

Or some might prefer to manage their services using simple yml files like [I did for testing v2](https://haste.antti.codes/fisibuqatu.yml)

TBlueF:
It tags them with the name of the branch .. meaning that you can't access them using bluemap's version? e.g. `3.0` .. 
Would that be possible to add?

Chicken:
Probably- but I have a skill issue and would like to not touch github actions for a while. I can PR that in the future.

PR #324, opened 2022-08-22
webserver.conf: Restore "ip" configuration value in the default configuration template that was erroneously removed during the configuration system refactor.

akemin-dayo:
webserver.conf: Restore "ip" configuration value in the default configuration template that was erroneously removed during the configuration system refactor in commit e555d558b7b9d93d4c10f30bec9922fce272c5b3.

This allows users to bind the BlueMap webserver to `127.0.0.1` again, which is useful for reverse proxy setups.

### DIFF
--- a/BlueMapCommon/src/main/resources/de/bluecolored/bluemap/config/webserver.conf
+++ b/BlueMapCommon/src/main/resources/de/bluecolored/bluemap/config/webserver.conf
@@ -13,6 +13,13 @@ enabled: true
 # Default is "bluemap/web"
 webroot: "${webroot}"
 
+# The IP address that the webserver will bind to.
+# The default value of "0.0.0.0" will bind to all available addresses, allowing the webserver to be accessed by any other machine.
+# If you want the webserver to ONLY be able to be accessed from your local machine, use "127.0.0.1" or "localhost".
+# This is useful if you want to put the BlueMap webserver behind a reverse proxy, such as lighttpd or nginx.
+# Default is "0.0.0.0"
+ip: "0.0.0.0"
+
 # The port that the webserver listens to.
 # Default is 8100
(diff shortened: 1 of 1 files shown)


Chicken:
This was an intentional decision to remove it from the template as it was often wrongly used by inexperienced people even with a warning.

akemin-dayo:
Ah, I see.

In that case, how do you and the other maintainers feel about perhaps putting a mention of it somewhere in [the documentation](https://github.com/BlueMap-Minecraft/BlueMapWiki), maybe in the reverse proxies section? (Speaking of, I should add lighttpd configuration instructions to the wiki, since it seems that no one else has done so…)

Personally speaking, when I was transitioning to BlueMap 3.3 (from 1.7.3) and was looking over the config file diffs (as I like to do when updating any of my plugins), the removal of the `ip` option from `webserver.conf` actually led me to erroneously believe that the functionality was completely removed _altogether_ (I was actually going to submit an issue about it). I think other server admins will likely run into this same misconception when upgrading, too.

As for me, I only realised that the functionality was actually still present when I decided to look through the source to find the webserver initialisation code, realising that it seemed like the config value was still being checked, and then just deciding to try adding the config value back in just to see what would happen.

Chicken:
Adding it to the wiki would indeed be the wisest choice.

TBlueF:
It indeed got intentionally hidden, because one of the most problems we had to solve on the support-discord, was because users just wouldn't stop ignoring all warnings and adding their `ip` to the config whilst not fully understanding their network-setup.

I have added an info box about this in the Reverse-Proxy wiki page :)
For everything else, we have a nice support-discord. If someone really misses this setting, they can always ask there about it :)

Nevertheless, thanks for your efforts on this PR! <3

PR #355, opened 2022-10-23
Add more style options to POI markers

Mark-225:
This PR adds two new properties to POI markers.

A 'scale' property to overwrite the size of the icon. This is useful when working with dynamic or user-generated icons that can be of different sizes but wanting to keep them all one consistent size in the browser.

An 'iconFilter' properly that gets set as the CSS 'filter' property on the  element. This allows for some basic post-processing on the client side rather than having to save each variant of an icon separately on the server-side.


TBlueF:
Closing this since it is now possible to add this using custom styles and css-classes for POI-markers :)

PR #390, opened 2023-02-08
Update finnish translation

Chicken:
Just in time


eemunni (approved):
Ye looks good. Ship it.

PR #414, opened 2023-04-21
Add version flag for cli

Chicken:
What kind of cli program can not print its own version?
Needed to rename mc-version flag because yeah... stupid flags.
I'm not sure if I like GitHub Codespaces anymore. Felt so sluggish today. Wrote this in a Dunkin' Donuts in the Cologne Central Station waiting for a damn train because everything is late or cancelled because of a hecking strike.


ankit-gautam23 (approved):
LGTM

PR #418, opened 2023-04-29
feat: Linear region file format support (extended from #415)

TBlueF:
Continuation of #415 (since i couldn't push to that branch ^^)..

@kugge i tried to make it a bit more extensible..
If you could re-test this to see if the linear implementation still works as intended, that would be great!


sofianedjerbi:
Hello. It works fine on my test server with `.linear` region file format.

PR #424, opened 2023-05-05
Linear timestamps

sofianedjerbi:
[Linear v2](https://github.com/KaiijuMC/Kaiiju/commit/3d7011f7d7077c4cb54e7220059cf299e002d989) added timestamps to allow software such as BlueMap to be more performant

**Offtopic:**
Chunk timestamps in minecraft only describe when the chunk is cleared or saved, so bluemap render chunks that haven't changed even when using mca region file format. 🤔


sofianedjerbi:
This is working as expected.

g0ldyy:
Hey, 
any chance you can update your PR so it works with the new bluemap regions interface 🙏

Issue #428, opened 2023-05-07
BlueMap threading error (stacktrace)


sofianedjerbi:
### What i did / Steps to reproduce
It happens randomly on player login.

### Expected result
No stacktrace.

### Actual result
Stacktrace:
```
[23:55:00 ERROR]: Could not pass event PlayerQuitEvent to BlueMap v3.13-13-dirty
java.util.ConcurrentModificationException: null
        at java.util.ArrayList.removeIf(ArrayList.java:1695) ~[?:?]
        at java.util.ArrayList.removeIf(ArrayList.java:1660) ~[?:?]
        at de.bluecolored.bluemap.bukkit.BukkitPlugin.onPlayerLeave(BukkitPlugin.java:254) ~[BlueMap-3.13-13-dirty-folia.jar:?]
        at com.destroystokyo.paper.event.executor.asm.generated.GeneratedEventExecutor20.execute(Unknown Source) ~[?:?]
        at org.bukkit.plugin.EventExecutor$2.execute(EventExecutor.java:77) ~[kaiiju-api-1.19.4-R0.1-SNAPSHOT.jar:?]
        at co.aikar.timings.TimedEventExecutor.execute(TimedEventExecutor.java:77) ~[kaiiju-api-1.19.4-R0.1-SNAPSHOT.jar:git-Kaiiju-"57bd665"]
        at org.bukkit.plugin.RegisteredListener.callEvent(RegisteredListener.java:70) ~[kaiiju-api-1.19.4-R0.1-SNAPSHOT.jar:?]
        at io.papermc.paper.plugin.manager.PaperEventManager.callEvent(PaperEventManager.java:54) ~[kaiiju-1.19.4.jar:git-Kaiiju-"57bd665"]
        at io.papermc.paper.plugin.manager.PaperPluginManagerImpl.callEvent(PaperPluginManagerImpl.java:126) ~[kaiiju-1.19.4.jar:git-Kaiiju-"57bd665"]
        at org.bukkit.plugin.SimplePluginManager.callEvent(SimplePluginManager.java:615) ~[kaiiju-api-1.19.4-R0.1-SNAPSHOT.jar:?]
        at net.minecraft.server.players.PlayerList.remove(PlayerList.java:663) ~[kaiiju-1.19.4.jar:git-Kaiiju-"57bd665"]
[...]

PR #429, opened 2023-05-07
Folia threading fix (#428)

sofianedjerbi:
A fix suggestion to #428


TBlueF:
Not quite sure why the change from `Collection` to `Set` is necessary..

However i am also very unsure what the f** i did in line 275, i don't hink I actually want to add a new `WeakReference<Task>` every second to that collection ...
I'll need to look over this myself again, this doesn't look right at all :D

PR #433, opened 2023-05-17
Update sql.conf

Modog500000:
Add
"# You still need to change your file type to sql in maps!" to sql.conf

### DIFF
--- a/BlueMapCommon/src/main/resources/de/bluecolored/bluemap/config/storages/sql.conf
+++ b/BlueMapCommon/src/main/resources/de/bluecolored/bluemap/config/storages/sql.conf
@@ -6,7 +6,7 @@
 # The storage-type of this storage.
 # Depending on this setting, different config-entries are allowed/expected in this config file.
 # Don't change this value! (If you want a different storage-type, check out the other example-configs)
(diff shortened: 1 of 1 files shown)


BlueMapCommon/src/main/resources/de/bluecolored/bluemap/config/storages/sql.conf
```diff
@@ -6,6 +6,7 @@
 # The storage-type of this storage.
 # Depending on this setting, different config-entries are allowed/expected in this config file.
 # Don't change this value! (If you want a different storage-type, check out the other example-configs)
+# You still need to change your file type to sql in maps!
```
Chicken:
```suggestion
# You still need to change your storage type to sql in your map configs!
```

Chicken (changes requested):

TBlueF:
A comment on the one option of the config-file that shouldn't be touched, doesn't seem to be placed very well to me tbh :D
I'll think about a better place for this :)

TBlueF:
Updated the wiki with a new page that explains the whole storage thing a bit more. This should be enough. :)

Issue #335, opened 2022-09-01
PostgreSQL SQL support


kloenk:
**Is your feature request related to a problem? Please describe.**
Tried setting BlueMap up using PostgreSQL, using the documentation for SQL.

**Describe the solution you'd like**
A sql syntax error is thrown by the PSQL jdbc adapter.

<details>
<summary>Stacktrace</summary>
<pre><code>
################################
 There is a problem with your BlueMap setup!
 Failed to load and initialize the storage 'sql'!
################################
[21:35:18] [Thread-15/ERROR]: Failed to load bluemap!
java.io.IOException: de.bluecolored.bluemap.common.config.ConfigurationException: java.io.IOException: org.postgresql.util.PSQLException: ERROR: syntax error at or near "`"
  Position: 28
	at de.bluecolored.bluemap.common.plugin.Plugin.load(Plugin.java:326) ~[cvav31clzs97zqn5afbjbr8mjpjbi6k5-BlueMap-3.3-fabric-1.19.jar:?]
	at de.bluecolored.bluemap.fabric.FabricMod.lambda$onInitialize$2(FabricMod.java:104) ~[cvav31clzs97zqn5afbjbr8mjpjbi6k5-BlueMap-3.3-fabric-1.19.jar:?]
	at java.lang.Thread.run(Thread.java:833) [?:?]
Caused by: de.bluecolored.bluemap.common.config.ConfigurationException: java.io.IOException: org.postgresql.util.PSQLException: ERROR: syntax error at or near "`"
  Position: 28
	at de.bluecolored.bluemap.common.BlueMapService.getStorage(BlueMapService.java:275) ~[cvav31clzs97zqn5afbjbr8mjpjbi6k5-BlueMap-3.3-fabric-1.19.jar:?]
	at de.bluecolored.bluemap.common.BlueMapService.loadWorldsAndMaps(BlueMapService.java:214) ~[cvav31clzs97zqn5afbjbr8mjpjbi6k5-BlueMap-3.3-fabric-1.19.jar:?]
	at de.bluecolored.bluemap.common.BlueMapService.getWorlds(BlueMapService.java:158) ~[cvav31clzs97zqn5afbjbr8mjpjbi6k5-BlueMap-3.3-fabric-1.19.jar:?]
[...]

PR #443, opened 2023-06-12
Feature : support for postgres databases

mrsolarius:
This pull request implements changes that make the usage of Postgres databases possible. It responds to this issue #335.

To achieve this, I have made the following improvements :
- Implemented an abstract factory to provide SQL queries that match the selected dialect.
- Created an abstract class called SQLStorage and two new classes that implement requests that cannot be implemented in the same way for Postgres and MySQL.
- Modified the storage type to utilize the static method create of SQLStorage, which parses the connection string to obtain a concrete instance of SQLStorage.

This is my first open-source pull request, so if I have made any mistakes, please let me know. Your feedback is greatly appreciated!


TBlueF:
I added just a couple of renames (because i am especially picky) and i got rid of the tiny bit of reflection you sneaked in there :3

PR #477, opened 2023-09-24
Add map config for having a different void colour

TechnicJelle:
Still to fix:
- [x] When the world actually has a void (like in the End), it's black, and not sky coloured.
	- [x] Figure out where the colour is coming from.
- [x] Doesn't work when Night Mode is enabled.
	- [x] Include the ambient/block/sky light calculations in the skyColor.
- [x] Change isVoid toggle to proper voidColor setting instead.
- [x] Add map config option for this.
	- [x] Research the `skyColor` option that already exists.

Demonstration videos:

https://github.com/BlueMap-Minecraft/BlueMap/assets/22576047/2bedbeb0-e973-440f-9f03-0ff843cc9488

https://github.com/BlueMap-Minecraft/BlueMap/assets/22576047/55d9cb53-1cab-4d86-9c77-cf0ef5d7adc0


TBlueF (approved):
Love it!

PR #481, opened 2023-09-30
Migrate fabric implementation to events system

NikitaCartes:
Instead of using mixins, it's possible to use join/disconnect events from fabric-api.
This makes it easier to develop for new versions.
There is also no longer a need for version 1.20.2, as it is now supported by the 1.20 version.


TBlueF:
Nice!
Only one concern:

The join/leave events are also used to track the players that are currently online (using a List and a Map). 
On the event that a player that is already on the server but logs in again (e.g. they crashed but didn't timeout yet on the server or log in from a different location) it has to be guaranteed that the leave-event always triggers before the join event.
Has this been tested already? :)

NikitaCartes:
Tested in different scenarios with both systems at the same time except for timeout (don't know how).
In all cases, the disconnect event is called first, then the join event. Same with log in from another location.

```
[03:52:16] [User Authenticator #2/INFO] (Minecraft) UUID of player NikitaCartes is 963df716-8938-4beb-bc3c-93db637e0625
[03:52:16] [Server thread/INFO] (Minecraft) NikitaCartes[/127.0.0.1:63498] logged in with entity id 238 at (-249.5, 80.0, 342.5)
[03:52:16] [Server thread/INFO] (Minecraft) [STDOUT]: Fabric event join
[03:52:16] [Server thread/INFO] (Minecraft) NikitaCartes joined the game
[03:52:16] [Server thread/INFO] (Minecraft) [STDOUT]: Bluemap event join
[03:52:23] [User Authenticator #3/INFO] (Minecraft) UUID of player NikitaCartes is 963df716-8938-4beb-bc3c-93db637e0625
[03:52:23] [Netty Server IO #2/INFO] (Minecraft) [STDOUT]: Fabric event disconnect
[03:52:23] [Server thread/INFO] (Minecraft) NikitaCartes lost connection: You logged in from another location
[03:52:23] [Server thread/INFO] (Minecraft) NikitaCartes left the game
[03:52:23] [Server thread/INFO] (Minecraft) [STDOUT]: Bluemap event leave
[03:52:23] [Server thread/INFO] (Minecraft) NikitaCartes[/127.0.0.1:63504] logged in with entity id 268 at (-249.5, 80.0, 342.5)
[03:52:23] [Server thread/INFO] (Minecraft) [STDOUT]: Fabric event join
[03:52:23] [Server thread/INFO] (Minecraft) NikitaCartes joined the game
[03:52:23] [Server thread/INFO] (Minecraft) [STDOUT]: Bluemap event join
[03:52:30] [Netty Server IO #3/INFO] (Minecraft) [STDOUT]: Fabric event disconnect
[03:52:30] [Server thread/INFO] (Minecraft) NikitaCartes lost connection: Disconnected
[03:52:30] [Server thread/INFO] (Minecraft) NikitaCartes left the game
[03:52:30] [Server thread/INFO] (Minecraft) [STDOUT]: Bluemap event leave
```

The main difference is that Fabric event is triggered when the [`onPlayerConnect`](https://github.com/NikitaCartes-archive/MinecraftDeobfuscated-Yarn/blob/5aa7b836dadebe1972f93bf91430aa4485aa2961/net/minecraft/server/PlayerManager.java#L147) function is called, while the original event is triggered when that function is completed. Although I didn't notice any side effects because of this.

TBlueF:
Thanks a lot for taking the time to test this! :)

PR #486, opened 2023-10-11
PDO php script implementation

AlbusRex:
I have tried to rewrite the current mysql.php to a more generic implementation using php's PDO features and I think I have succeeded.
This implementation has been tested with NGINX and a Postgre storage backend and has been accessed by chromium and firefox based browsers.
On the backend getting PDO to work is very easy. Just by installing the corresponding php database module the PDO driver will become available.
I have not had the opportunity to test a different database storage unfortunately.


BlueMapCommon/webapp/public/map_data_provider.php
```diff
@@ -126,67 +128,72 @@ function getMimeType($path) {
...
+                    header("Content-Type: image/png");
+                }
+                fpassthru($line["data"]);
+                exit;
             }
 
-            echo $line["data"];
-            exit;
-        }
+        } catch (PDOException $e) { error(500, "$e"); }
```
TBlueF: Which error-messages are being outputted here, and can we be sure that this is nothing which could be dangerous to send to the public? (E.g. backend-ip's/adresses, users etc..)

BlueMapCommon/webapp/public/map_data_provider.php
```diff
@@ -2,19 +2,18 @@
...
 // !!! SET YOUR SQL-CONNECTION SETTINGS HERE: !!!
 
+$driver   = 'pgsql'; //PDO driver name, check the PDO documentation for available drivers at https://www.php.net/manual/en/pdo.drivers.php. Common ones are 'pgsql' or 'mysql'.
 $hostname = '127.0.0.1';
-$port     = 3306;
-$username = 'root';
+$port     = 5432; // Remember to change this value to match your database configuration
+$username = '';
 $password = '';
 $database = 'bluemap';
```
TBlueF: I'd like to keep the default at the mysql settings that we had before, since (afaik) more people are using mysql than postgresql :)

BlueMapCommon/webapp/public/map_data_provider.php
```diff
@@ -126,67 +128,72 @@ function getMimeType($path) {
...
+        $statement->execute();
+
+        if ($line = $statement->fetch()) {
         header("Content-Type: ".getMimeType($mapPath));
-        echo $line["value"];
+        fpassthru($line["value"]);
         exit;
-    }
+        }
+    } catch (PDOException $e) { error(500, "$e"); }
```
TBlueF: Which error-messages are being outputted here, and can we be sure that this is nothing which could be dangerous to send to the public? (E.g. backend-ip's/adresses, users etc..)

BlueMapCommon/webapp/public/map_data_provider.php
```diff
@@ -111,9 +110,12 @@ function getMimeType($path) {
...
     $mapPath = explode("?", $pathParts[1], 2)[0];
 
-    // get sql-connection
-    $sql = new mysqli($hostname, $username, $password, $database, $port);
-    if ($sql->errno) error(500, "Failed to connect to Database!");
+    // Initialize PDO
+    try {
+        $sql = new PDO("$driver:host=$hostname;dbname=$database", $username, $password);
+        $sql->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
+    } catch (PDOException $e ) { error(500, "$e"); }
```
TBlueF: Which error-messages are being outputted here, and can we be sure that this is nothing which could be dangerous to send to the public? (E.g. backend-ip's/adresses, users etc..)

TBlueF (changes requested):
Looks very nice! Thank you for the efforts so far :)

I'll still need to test this on my end, but here are some things i noticed when reading over this quickly..

Additionally to the comments, i think naming the file `sql.php` instead of `map_data_provider.php` would be easier and more understandable for many users. If you agree, could you please rename the file to that instead? :)

BlueMapCommon/webapp/public/map_data_provider.php
AlbusRex: That returns a very detailed PDO exception that details what went wrong.
It is indeed possible to leak info, I will change it to a generic message.
There is also the possibility of setting a "debug=true" somewhere in the future to output detailed information.

BlueMapCommon/webapp/public/map_data_provider.php
AlbusRex: Seems like a good idea

AlbusRex (commented):
I have implemented the changes that were suggested.
I am a bit slow on the update since I am new to this workflow on github.

TBlueF:
Thanks for the changes, looks good now :)
I'll test it as soon as i find some time and if everything works as expected it will be merged 👍

TBlueF (approved):

TBlueF:
@AlbusRex Sorry for the very long time it took me to test this..

I had to change [a couple of things](https://github.com/BlueMap-Minecraft/BlueMap/pull/486/commits/51f0c781025eed5dd0e3cf81daca799c8461325d) to make this work on my end with mysql.
Can you please confirm that my changes didn't break anything, and it still works with PostgreSQL ? :)

AlbusRex:
I have tested the changes and everything works as intended. I suspect we had one of those "by value vs by reference" kind of problems.

PR #511, opened 2024-02-23
b2 support

vgskye:
Adds storage backend for Backblaze B2 API.


vgskye:
#512 supports a strict superset of this(B2 provides a S3-compatible API as well), closing

PR #512, opened 2024-02-23
S3 support

vgskye:
Adds support for using S3(and other compatible providers) as a storage backend.

### DIFF
--- a/BlueMapCommon/src/main/java/de/bluecolored/bluemap/common/config/storage/S3Config.java
+++ b/BlueMapCommon/src/main/java/de/bluecolored/bluemap/common/config/storage/S3Config.java
@@ -8,15 +8,14 @@ import org.spongepowered.configurate.objectmapping.ConfigSerializable;
 import java.util.Optional;
 
 @SuppressWarnings({"FieldMayBeFinal", "FieldCanBeLocal"})
-@DebugDump
 @ConfigSerializable
 public class S3Config extends StorageConfig implements S3StorageSettings {
-    private String endpoint = null;
-    private String region = "";
-    private String accessKey = "";
+    @DebugDump private String endpoint = null;
+    @DebugDump private String region = "";
+    @DebugDump private String accessKey = "";
     private String secretKey = "";
-    private String bucket = "";
-    private Compression compression = Compression.GZIP;
+    @DebugDump private String bucket = "";
+    @DebugDump private Compression compression = Compression.GZIP;
 
     @Override
     public Optional<String> getEndpoint() {
--- a/BlueMapCommon/src/main/java/de/bluecolored/bluemap/common/config/storage/S3Config.java
+++ b/BlueMapCommon/src/main/java/de/bluecolored/bluemap/common/config/storage/S3Config.java
@@ -1,0 +1,50 @@
+package de.bluecolored.bluemap.common.config.storage;
+
+import de.bluecolored.bluemap.api.debug.DebugDump;
+import de.bluecolored.bluemap.core.storage.Compression;
+import de.bluecolored.bluemap.core.storage.s3.S3StorageSettings;
+import org.spongepowered.configurate.objectmapping.ConfigSerializable;
+
+import java.util.Optional;
+
+@SuppressWarnings({"FieldMayBeFinal", "FieldCanBeLocal"})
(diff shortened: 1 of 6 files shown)


BlueMapCommon/src/main/java/de/bluecolored/bluemap/common/config/storage/S3Config.java
```diff
@@ -0,0 +1,50 @@
...
+
+import de.bluecolored.bluemap.api.debug.DebugDump;
+import de.bluecolored.bluemap.core.storage.Compression;
+import de.bluecolored.bluemap.core.storage.s3.S3StorageSettings;
+import org.spongepowered.configurate.objectmapping.ConfigSerializable;
+
+import java.util.Optional;
+
+@SuppressWarnings({"FieldMayBeFinal", "FieldCanBeLocal"})
+@DebugDump
```
TBlueF: We can't debug-dump all values of this config, since it contains secrets. Only dump the non-secret ones :)

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/storage/s3/S3Storage.java
```diff
@@ -0,0 +1,342 @@
...
+                            .key(path)
+                            .build()
+            );
+            return Optional.of(new CompressedInputStream(is, compression));
+        } catch (Exception e) {
+            if (e instanceof NoSuchKeyException) {
+                return Optional.empty();
+            }
+            throw new IOException(e);
+        }
```
TBlueF:
```suggestion
        } catch (NoSuchKeyException e) {
            return Optional.empty();
        } catch (SdkException e) {
            throw new IOException(e);
        }
```
Same for all the other catches :)

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/storage/s3/S3Storage.java
```diff
@@ -0,0 +1,342 @@
...
+                            .bucket(bucket)
+                            .key(path)
+                            .build()
+            );
+        } catch (Exception e) {
+            if (e instanceof NoSuchKeyException) {
+                return;
+            }
+            throw new IOException(e);
+        }
```
TBlueF:
```suggestion
        } catch (NoSuchKeyException ignore) {
        } catch (SdkException e) {
            throw new IOException(e);
        }
```

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/storage/s3/S3Storage.java
```diff
@@ -0,0 +1,342 @@
...
+                                .build()
+                );
+
+                if (!onProgress.apply(
+                        new ProgressInfo((double) (i + 1) / filesList.size())
+                )) return;
+            }
+        } catch (Exception e) {
+            if (e instanceof NoSuchKeyException) {
+                return;
```
TBlueF: Can we really ignore this here? Since it interrupts the deletion-process, it should either throw an error, or better catched earlier so it doesn't interrupt the rest of the deletions.

TBlueF (commented):
Thank you for the PR!
As discussed in Discord already, the Problem with it is that the aws-sdk increases BlueMap's jar size by too much.
So we will probably wait until its possible to make an addon for bluemap that adds a new storage-type and then separate this PR into such an addon :)

I still went ahead and commented all the thoughts on the code that i had found right now :)

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/storage/s3/S3Storage.java
vgskye: remnant from me using minio at first and switching to aws-sdk only after it failed to build, yeah

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/storage/s3/S3Storage.java
vgskye: yeah there should be a try-catch in the loop body there

vgskye:
(force push is just a rebase)

PR #513, opened 2024-02-24
S3 support, take 2

vgskye:
replaces #512. Adds S3 support using my fork of s3-lite to be significantly lighter than aws-sdk, weighing at 100KiB-or-so when using URLConnection and about 1.5MB-or-so when using the supposedly higher-performance Apache HttpComponents client.

### DIFF
--- a/BlueMapCommon/src/main/java/de/bluecolored/bluemap/common/config/storage/S3Config.java
+++ b/BlueMapCommon/src/main/java/de/bluecolored/bluemap/common/config/storage/S3Config.java
@@ -8,15 +8,14 @@ import org.spongepowered.configurate.objectmapping.ConfigSerializable;
 import java.util.Optional;
 
 @SuppressWarnings({"FieldMayBeFinal", "FieldCanBeLocal"})
-@DebugDump
 @ConfigSerializable
 public class S3Config extends StorageConfig implements S3StorageSettings {
-    private String endpoint = null;
-    private String region = "";
-    private String accessKey = "";
+    @DebugDump private String endpoint = null;
+    @DebugDump private String region = "";
+    @DebugDump private String accessKey = "";
     private String secretKey = "";
-    private String bucket = "";
-    private Compression compression = Compression.GZIP;
+    @DebugDump private String bucket = "";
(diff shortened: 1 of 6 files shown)


vgskye:
apparently something's broken, sec

vgskye:
yeah s3-lite's signer seems broken

vgskye:
all fixed now, seems to be perfectly functional

TBlueF:
So ...

BlueMap v5.0 (Snapshot) now supports not only native-addons (addons loaded by bluemap), but it is also possible to register new  storages or compression etc. via an addon that depends on BlueMapCore/Common.
With that it should be possible to create a plugin that adds S3-Storage support to BlueMap :)

For example here is an addon that adds brotli-support to bluemap: https://github.com/BlueMap-Minecraft/BlueMapBrotli

For S3, you could implement the [`Storage`](https://github.com/BlueMap-Minecraft/BlueMap/blob/master/BlueMapCore/src/main/java/de/bluecolored/bluemap/core/storage/Storage.java) interface in your addon, and then register it to the [`StorageType.REGISTRY`](https://github.com/BlueMap-Minecraft/BlueMap/blob/master/BlueMapCommon/src/main/java/de/bluecolored/bluemap/common/config/storage/StorageType.java#L38).

I didn't test it but in theory it should be even possible using a library like: https://github.com/elerch/Amazon-S3-FileSystem-NIO2 or https://github.com/carlspring/s3fs-nio/ to aquire a `java.nio.file.Path` object and then use that to simply create a [`new FileStorage(root, compression, false)`](https://github.com/BlueMap-Minecraft/BlueMap/blob/master/BlueMapCore/src/main/java/de/bluecolored/bluemap/core/storage/file/FileStorage.java#L42).. If that works, you wouldn't even need to implement the `Storage` interface at all.

BlueMap 5.0 is still a snapshot, so it's not well tested yet.. but that also means that if you encounter any issues, just tell me and the path to a fix should usually be rather short :)

I will close this PR, since as we already discussed on discord, implementing this as an addon instead of in BlueMap directly is the preferred way :)

PR #515, opened 2024-02-25
Update ru translation

NikitaCartes:
Also: shouldn't `blockTooltip.light` be "sky" instead of "sun" to match in-game f3 menu?


TBlueF:
I have no idea, probably ... there is no consistency at all, in some places it's called sky-light and in others sunlight :D

PR #524, opened 2024-03-17
Add mechanism for retrieving BlockEntity data

glorantq:
This commit adds a mechanism for retrieving block entity data. Block entity data is required to support for example text on signs, banner patterns, or mods such as Domum Ornamentum.

### DIFF
--- a/BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/block/Block.java
+++ b/BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/block/Block.java
@@ -28,6 +28,8 @@ import de.bluecolored.bluemap.core.world.BlockState;
 import de.bluecolored.bluemap.core.world.Chunk;
 import de.bluecolored.bluemap.core.world.LightData;
 import de.bluecolored.bluemap.core.world.World;
+import de.bluecolored.bluemap.core.world.block.entity.BlockEntity;
+import org.jetbrains.annotations.Nullable;
 
 public class Block<T extends Block<T>> {
 
@@ -147,6 +149,10 @@ public class Block<T extends Block<T>> {
         return getLightData().getBlockLight();
     }
 
+    public @Nullable BlockEntity getBlockEntity() {
+        return getChunk().getBlockEntity(x, y, z);
+    }
+
     @Override
     public String toString() {
         if (world != null) {
@@ -174,5 +180,4 @@ public class Block<T extends Block<T>> {
     protected T self() {
         return (T) this;
     }
-
 }
--- a/BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/block/entity/BlockEntity.java
+++ b/BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/block/entity/BlockEntity.java
@@ -13,15 +13,16 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.StringJoiner;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 @NBTDeserializer(BlockEntity.BlockEntityDeserializer.class)
 public class BlockEntity {
     private static final BlueNBT BLUENBT = new BlueNBT();
-    private static final Map<String, Class<? extends BlockEntity>> ID_MAPPING = Map.of(
-            "minecraft:sign", SignBlockEntity.class,
-            "minecraft:skull", SkullBlockEntity.class,
-            "minecraft:banner", BannerBlockEntity.class
+    private static final Map<String, Function<Map<String, Object>, ? extends BlockEntity>> ID_MAPPING = Map.of(
+            "minecraft:sign", SignBlockEntity::new,
+            "minecraft:skull", SkullBlockEntity::new,
+            "minecraft:banner", BannerBlockEntity::new
     );
 
     protected final String id;
@@ -91,16 +92,10 @@ public class BlockEntity {
                 return null;
             }
 
-            Class<? extends BlockEntity> dataClass = ID_MAPPING.getOrDefault(id, BlockEntity.class);
+            Function<Map<String, Object>, ? extends BlockEntity> instance = ID_MAPPING.getOrDefault(id, BlockEntity::new);
 
             try {
-                Constructor<? extends BlockEntity> constructor = dataClass.getDeclaredConstructor(Map.class);
-                if (constructor.trySetAccessible()) {
-                    return constructor.newInstance(data);
-                }
-            } catch (NoSuchMethodException e) {
-                Logger.global.logError(
-                        String.format("No constructor in %s that takes a Map!", dataClass.getCanonicalName()), e);
+                return instance.apply(data);
             } catch (Exception e) {
                 Logger.global.logError("Failed to instantiate BlockEntity instance!", e);
             }
--- a/BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/block/entity/BlockEntity.java
+++ b/BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/block/entity/BlockEntity.java
@@ -1,3 +1,27 @@
+/*
+ * This file is part of BlueMap, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) Blue (Lukas Rieger) <https://bluecolored.de>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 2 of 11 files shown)


BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/block/BlockEntity.java
```diff
@@ -0,0 +1,142 @@
...
+
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+public class BlockEntity {
+    private static final List<String> COMMON_KEYS = List.of("id", "x", "y", "z", "keepPacked");
+
+    private final String id;
```
TBlueF: this can be a `Key` instead of a `String`

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/block/BlockEntity.java
```diff
@@ -0,0 +1,142 @@
+package de.bluecolored.bluemap.core.world.block;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+public class BlockEntity {
+    private static final List<String> COMMON_KEYS = List.of("id", "x", "y", "z", "keepPacked");
```
TBlueF:
```suggestion
    private static final Set<String> COMMON_KEYS = Set.of("id", "x", "y", "z", "keepPacked");
```

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/block/Block.java
```diff
@@ -147,6 +149,10 @@ public int getBlockLightLevel() {
         return getLightData().getBlockLight();
     }
 
+    public Optional<BlockEntity> getBlockEntity() {
```
TBlueF: Using an `Optional` is definitely the better practice, but i like to just return a nullable `BlockEntity` + `@Nullable` annotation here, to micro-optimize the extra wrapper object away ^^'

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/block/BlockEntity.java
```diff
@@ -0,0 +1,142 @@
...
+        return tree.getOrDefault(key, def);
+    }
+
+    public Object get(String key, Object def) {
+        return get(data, key, def);
+    }
+
+    public String getString(String key, String def) {
+        return getString(data, key, def);
+    }
```
TBlueF: This doesn't feel right..
A lot of those methods could be `static`, which then begs the question why we need them in the first place, they look like a lenient but expensive way of casting any "object" into the value-type we need.

I understand the issue that you try to solve here: each tile-entity could have a different data-structure and thus can't be represented by one object-type with a fixed set of properties.

My approach for this would be to create a tree of fixed types anyways (e.g. `SignBlockEntity extends BlockEntity`) with well defined properties, and then write a [custom `TypeDeserializer<BlockEntity>`](https://github.com/BlueMap-Minecraft/BlueMap/tree/master/BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/mca/data) which first reads the whole data-tree (i don't think this can be avoided) but then (based on the `id`) creates the correct sub-type of `BlockEntity`.

Also i think instead of trying to convert between value-types (e.g. `Integer.parseInt(String.valueOf(obj))`) we should just cast based on the type that we expect (`(int) obj`) and deal with the cast-exception and treat it as invalid data-format.

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/mca/chunk/Chunk_1_13.java
```diff
@@ -195,6 +204,13 @@ public int getOceanFloorY(int x, int z) {
         );
     }
 
+    @Override
+    public Optional<BlockEntity> getBlockEntity(int x, int y, int z) {
+        return blockEntities.stream()
+                .filter(it -> it.getX() == x && it.getY() == y && it.getZ() == z)
+                .findFirst();
+    }
```
TBlueF: (return `@Nullable BlockEntity`, see comment above)

And we should definitely try to avoid any type of object-creation here (e.g. streams) and consider using some kind of hash-lookup or array-index to not have to loop over **all** tile-entities.
People have been filling chunks with flower-pot's already and tried to render them with bluemap so i can well imagine someone having hundreds of signs as well ^^

Maybe just splitting them into the individual chunk-sections could be a good-enough approach already 🤔

TBlueF (changes requested):
Thanks for the PR!
Just a really quick and hasty review for now, since i don't have much time rn until tuesday :)

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/mca/chunk/Chunk_1_13.java
```diff
@@ -273,6 +289,7 @@ public static class Level {
         private HeightmapsData heightmaps = new HeightmapsData();
         private SectionData @Nullable [] sections = null;
         private int[] biomes = EMPTY_INT_ARRAY;
+        @NBTName("TileEntities") private List<Map<String, Object>> blockEntities = List.of(Map.of());
```
TBlueF: With the `TypeDeserializer<BlockEntity>` mentioned above, this can then be simplified into a `List<BlockEntity>` :)

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/block/Block.java
glorantq: Done

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/block/BlockEntity.java
glorantq: Not needed with the class-hierarchy approach

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/block/BlockEntity.java
glorantq: Not needed with the class-hierarchy approach

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/mca/chunk/Chunk_1_13.java
glorantq: Done

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/block/BlockEntity.java
glorantq: Done

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/mca/chunk/Chunk_1_13.java
glorantq: Done, replaced with packing the coordinates into a long and using that as a key for a Map. The only way to make this faster would be to make a custom map implementation that eliminates the small impact of the boxing, or bring over one of the `Long2Object*` maps from fastutil.

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/block/entity/BlockEntity.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.StringJoiner;
+import java.util.stream.Collectors;
+
+@NBTDeserializer(BlockEntity.BlockEntityDeserializer.class)
+public class BlockEntity {
+    private static final BlueNBT BLUENBT = new BlueNBT();
+    private static final Map<String, Class<? extends BlockEntity>> ID_MAPPING = Map.of(
```
TBlueF: How about switching this for a `Map<String, Function<Map<String, Object>, ? extends BlockEntity>>` instead?
Then we can store the type's constructor in the map directly and don't need to use reflection below :)

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/mca/chunk/Chunk_1_13.java
```diff
@@ -113,6 +119,10 @@ public Chunk_1_13(MCAWorld world, Data data) {
             this.sectionMin = 0;
             this.sectionMax = 0;
         }
+
+        this.blockEntities = level.blockEntities.stream().collect(Collectors.toMap(
+                it -> (long) it.getY() << 40 | (it.getX() & 0xF) << 4 | it.getZ() & 0xF, it -> it
```
TBlueF: I really like this solution for quick access!
Can you explain why y is shifted by `40` tho? I would have expected `8` or maybe `32` 🤔

glorantq: You are right, it should be 8. It was late, the 40 came from 8 (x and z) + 32 (y) because for some reason I thought there that the shift moves the most significant bit. Oops.

BlueMapCore/src/main/java/de/bluecolored/bluemap/core/world/block/entity/BlockEntity.java
glorantq: Right, apart from the ungodly type we can just store a method reference to the constructor

TBlueF: Can always write a small functional interface to make the type more readable :D

```java
Map<String, BlockEntityConstructor> ID_MAPPING = Map.of(
     "minecraft:sign", SignBlockEntity::new,
     "minecraft:skull", SkullBlockEntity::new,
     "minecraft:banner", BannerBlockEntity::new
);

@FunctionalInterface
private interface BlockEntityConstructor {
    BlockEntity construct(Map<String, Object> data);
}
```
super fancy :)

TBlueF:
The github-build fails, but i think it only needs a `./gradlew spotlessApply` to add the license headers ^^

glorantq:
Yeah I realised I missed the headers, should be fine now. `spotlessCheck` is happy locally.

TBlueF (approved):

Issue #145, opened 2021-01-31
Add resource-extensions for all missing tile-entities & special-blocks


ChloeDawn:
**Is your feature request related to a problem? Please describe.**
Banners are currently invisible due to them being dynamically rendered by the game

**Describe the solution you'd like**
A static model that represents a banner (can easily be made to look identical)

TBlueF:
Adding all the tile-entities that are still missing (banners, shulker-chests, the bell of the bell, mob-/player-heads) is already somewhere on the TODO :)
The bigger problem with banners is the banner-texture, since that is the same issue i mentioned on #146 ^^
That is also planned, but prioritized a little bit lower right now ^^

TBlueF:
Missing resources:
- [ ] banners
- [ ] shulker-chests
- [ ] bell
- [ ] mob/player-heads
- [ ] end-portal/end-gateway
- [ ] hanging-signs
- [ ] cherry-sign
- [ ] decorated pots
- [ ] cake-block (?)

ChloeDawn:
Personally I'd be fine with a blank banner model as a stop-gap to pattern handling. It would be better than no banner imo ^^

TyBraniff:
[Live Examples](https://bluemap.bronzedeathsquad.com/spigot/#flat:-5:-56:13:5:0.48:0.99:0:0:perspective)

Tonight after the kids head off to bed I will finish all heads/skulls and the decorated vases.
With that I think that'll be everything except for the end-portal which I have *no idea* how to do so y'all're on your own for that one

PR #536, opened 2024-05-01
Devolvement of issues #145 + missing signs

TyBraniff:
Added Cherry and Bamboo signs.


TyBraniff:
Updated to include everything from my fixes resource pack which should satisfy most of https://github.com/BlueMap-Minecraft/BlueMap/issues/145 

Not sure I'm doing these pull requests right, hopefully its all copasetic. 

[Live Examples](https://bluemap.bronzedeathsquad.com/spigot/#flat:-5:-56:13:5:0.48:0.99:0:0:perspective)

BlueMapCore/src/main/resourceExtensions/mc1_20_3/assets/minecraft/models/entity/decorated_pot.json
```diff
@@ -0,0 +1,48 @@
+{
+	"parent": "Made with Blockbench by TyBraniff for Bluemaps support.",
```
TBlueF: Pretty sure this was supposed to be in `"credit"` instead of `"parent"`, that might even be the reason why they are not working =)

Also a side-note on this: 
As much as i am all for giving credit to contributors, imo this is the wrong place ^^
We can keep the notes in there for now, so no need to go over all of them and remove it or something, but just so you know that they might be removed in the future for any reason :)

BlueMapCore/src/main/resourceExtensions/mc1_20_3/assets/minecraft/models/entity/dragon_head.json
```diff
@@ -0,0 +1,101 @@
+{
+	"parent": "Made with Blockbench by TyBraniff for Bluemaps support.",
```
TBlueF: ("parent" vs "credit", see previous comment)

BlueMapCore/src/main/resourceExtensions/mc1_20_3/assets/minecraft/models/entity/dragon_wall_head.json
```diff
@@ -0,0 +1,115 @@
+{
+	"parent": "Made with Blockbench by TyBraniff for Bluemaps support.",
```
TBlueF: ("parent" vs "credit", see previous comment)

TBlueF (changes requested):
Great work! Thanks a ton for this :)
And yes, you are doing these pull requests perfectly right :D

TyBraniff:
Was hopeful that would fix the decorated_pot not loading right. However it didn't. Edits are made. Thanks for the catch.

TBlueF:
(Squash-) Merged this into my v5 branch so github retains you as the commit-author :)
I will be restructuring the resource-extensions a bit, but like this you won't have to do anything and it will be included in BlueMap v5 which i am rn working towards :)

@TyBraniff Again, awesome work, thanks a lot for this PR!

Phoenix616:
All the non-full rotations of the hanging signs have a 0.5 offset to the true values :S (e.g. `22` should be `22.5` [here](https://github.com/BlueMap-Minecraft/BlueMap/pull/536/files#diff-83d0855d7edb9e384be37c008e81143a0afcbe1f4392fc74c18a33eba8fa56bbR40) as it's `45/2` or `360/16`)

TBlueF:
@Phoenix616 Good catch!
If you think the effort is worth it for half a degree, please PR a fix :)

PR #542, opened 2024-05-13
Show chunk borders

TechnicJelle:
image
image

Should probably add a toggle for it, though.
And a way to customise the colour.

### DIFF
--- a/BlueMapCommon/webapp/public/lang/en.conf
+++ b/BlueMapCommon/webapp/public/lang/en.conf
@@ -83,6 +83,9 @@
     light: "Light"
     contrast: "Contrast"
   }
+  chunkBorders: {
+    button: "Show chunk borders"
+  }
   debug: {
     button: "Debug"
   }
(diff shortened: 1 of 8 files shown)


TBlueF:
Great work!
I'll not merge this without a toggle, though ^^

TechnicJelle:
Aw dangit, I was hoping I wouldn't have to do that myself :p

TBlueF:
I mean .. you can also wait for me to do it, but then it'll end up on the todo and probably take a while to be prioritised ^^

TechnicJelle:
Fair enough...
I guess I'll see how horrific vue.js is, then :p

TBlueF:
If you move the show-chunk-borders variable away from the appState directly to the mapviewer and do it like the `sunlightStrength` and `ambientLight` sliders do it, then you don't need to extend the poor MapViewer constructor =)

And then i am happy :3

PR #570, opened 2024-07-10
updateMap will be automatically triggered on page reload 

MicLieg:
As suggested in this [Discord message](https://discord.com/channels/665868367416131594/751804128749027421/1259804226645921863), this PR adds a mounted lifecycle hook to the MainMenu component in order to update the map when the page is reloaded or refreshed using the [PerformanceNavigationTiming: type property](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming/type).

This will help users who are trying to troubleshoot missing parts of the map by refreshing the map instead of using the cached tiles.

Feedback on this implementation is greatly appreciated as I'm new to Vue.

### DIFF
--- a/BlueMapCommon/webapp/src/components/Menu/MainMenu.vue
+++ b/BlueMapCommon/webapp/src/components/Menu/MainMenu.vue
@@ -52,13 +52,6 @@ export default {
       markers: this.$bluemap.mapViewer.markers.data,
     }
   },
-  mounted() {
-    // Update the map when the page is reloaded / refreshed because it's probably a troubleshooting step
-    const [entry] = performance.getEntriesByType("navigation");
-    if (entry.type === "reload") {
-      this.$bluemap.updateMap();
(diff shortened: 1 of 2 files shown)


TBlueF (changes requested):
Thank you for taking the time to make a PR on this :)

My suggestion would be instead of adding this to the `MainMenu.vue` and "simulating" a button press, prevent the tile-cache-hash update when loading the webapp:

E.g. here: https://github.com/BlueMap-Minecraft/BlueMap/blob/master/BlueMapCommon/webapp/src/js/BlueMapApp.js#L601
the cache-hash used to request tiles is being loaded. So we could just surround that line into an `if(!reloadDetected) { ... }` to prevent the hash-loading and thus use a new cache-hash -> map-tiles get re-requested :)

TBlueF (approved):
perfect :)

Issue #398, opened 2023-03-11
Live daylight cycle


meyerpr:
Is it possible to add the server time to the map like in Dynmap? Maybe even an option to switch the map to day/night automatically.

Thanks

TBlueF:
This was already on the [TODO](https://github.com/orgs/BlueMap-Minecraft/projects/2) as a note .. i replaced the note with this issue

darthm1ke:
I would love to contribute, I am enjoying this map plugin so much, comming back after 10 years and seeing this improvement over dynmap makes me want to go back into miniecraft servering. I too want this feature, and I wonder if maybe I can help with it.

MennoGamed:
I see this is still on the "ToDo List", wich makes me a little sad

Xaner98:
Push! Would love to see this

PR #571, opened 2024-07-11
Support the live daylight cycle

Soulter:
## Introduction
Thank you for your amazing project. BlueMap helps me a lot!
Just noticed that the feature is in Ideas board, so I try to implement it. #398
These changes implement the feature of a live daylight cycle on BlueMap webpages. I leverage the Minecraft API to get the ambient darkness of all dimensions and add an HTTP handler to allow frontend to retrieve this data.

## Demo
| Night | Dawn | Day |
|--------|--------|--------|
| 8d02837c0ac024d74127f7997db35c64 | 0301e8c17bf7e7436630777f6564c382 | d79d4a90344fd588753191702535af61 | 

## Details

- For the `forge` and `neoforge` sides, I utilized `level.getSkyDarken()` to get ambient darkness data. This method returns an integer ranging from 0 to 11, where 0 means day and 11 means night.
- The `fabric` side is similar to the `forge` side; I used `world.getAmbientDarkness()`, which has the same effect as `getSkyDarken()`.
- For the `spigot` and `sponge` sides, I couldn't find a similar function, so I used ticks and some hardcoding to implement the feature.
- On the frontend, I added a switch button in the `SettingsMenu`. Users can toggle this button to turn the feature on or off. The frontend will request `/live/brightness.json` every 2 seconds (or every 3 seconds if the request fails) and adjust the `sunlightStrength` value accordingly. The data format is as follows:
  ```json
  {
      "brightness": [
          {
              "dimension": "overworld",
              "val": 3
          },
          {
              "dimension": "the_nether",
              "val": 0
          },
          {
              "dimension": "the_end",
              "val": 0
          }
      ]
  }
  ```
  

> *I figured out a way to calculate ambient darkness as shown below, but it requires too many computational resources, so I didn't implement it for the `spigot` and `sponge` sides.
> 

## Test coverage
I tested forge, fabric, paper, and sponge on versions 1.19-1.20, and the feature works well. I can make more tests if you think this PR would be helpful for you.

If you find any errors or shortcomings in these changes, I will fix them as soon as possible! And the branch is allow edits by maintainers, you can make any improvement.

### DIFF
--- a/BlueMapCommon/src/main/java/de/bluecolored/bluemap/common/serverinterface/Server.java
+++ b/BlueMapCommon/src/main/java/de/bluecolored/bluemap/common/serverinterface/Server.java
@@ -28,10 +28,12 @@ import de.bluecolored.bluemap.common.debug.DebugDump;
 import de.bluecolored.bluemap.core.util.Tristate;
 import de.bluecolored.bluemap.core.world.World;
 import de.bluecolored.bluemap.core.world.mca.MCAWorld;
+import de.bluecolored.bluemap.core.util.Key;
 import org.jetbrains.annotations.Nullable;
 
 import java.nio.file.Path;
 import java.util.Collection;
+import java.util.Map;
 import java.util.Optional;
 
 public interface Server {
@@ -98,6 +100,9 @@ public interface Server {
      */
     @DebugDump
     Collection<Player> getOnlinePlayers();
+
+    @DebugDump
+    Map<Key, Integer> getSkyBrightness();
 
     /**
      * Registers a ServerEventListener, every method of this interface should be called on the specified events
--- a/BlueMapCommon/webapp/src/js/BlueMapApp.js
+++ b/BlueMapCommon/webapp/src/js/BlueMapApp.js
@@ -97,7 +97,8 @@ export class BlueMapApp {
                 showZoomButtons: true,
                 invertMouse: false,
                 enableFreeFlight: false,
-                pauseTileLoading: false
+                pauseTileLoading: false,
+                liveDaylightCycle: false,
             },
             menu: this.mainMenu,
             maps: [],
@@ -120,6 +121,7 @@ export class BlueMapApp {
         this.mapViewer.markers.add(this.popupMarkerSet);
 
         this.updateLoop = null;
+        this.updateDaylightCycle = null;
 
         this.hashUpdateTimeout = null;
         this.viewAnimation = null;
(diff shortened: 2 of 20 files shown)


BlueMapCommon/webapp/src/js/BlueMapApp.js
```diff
@@ -383,6 +405,34 @@ export class BlueMapApp {
...
+        */
+        return new Promise((resolve, reject) => {
+            let loader = new FileLoader();
+            loader.setResponseType("json");
+            loader.load(this.mapViewer.map.data.dataUrl + "live/brightness.json?" + generateCacheHash(),
+                fileData => {
+                    if (!fileData || !fileData.brightness) {
+                        reject("Failed to parse brightness.json");
+                    }
+                    let world_name = this.mapViewer.map.data.name.match(/\(([^)]+)\)/)[1].toLowerCase();
```
TBlueF, line 420: The map-name can be changed and is not at all guaranteed to include the world-id/dimension, so we can't use this to identify the dimension-key of a map

BlueMapCommon/src/main/java/de/bluecolored/bluemap/common/serverinterface/Server.java
```diff
@@ -99,6 +101,9 @@ default Optional<ServerWorld> getServerWorld(Object world) {
     @DebugDump
     Collection<Player> getOnlinePlayers();
 
+    @DebugDump
+    Map<Key, Integer> getSkyBrightness();
```
TBlueF, line 105: I think this should be something like a `int getSkyBrightness()` or maybe even `int getWorldTime()` in the `ServerWorld` interface instead

TBlueF (commented):
Thanks a lot for taking the time to make a PR!

Unfortunately you didn't speak to me about this before implementing it, because i don't think bluemap is ready for this feature yet.. let me explain:

What i really don't like about this is the need to have yet another additional constant request-spam to the webserver for each update.
This is already something i want to change about the player/marker updates.

The plan is to eventually switch to a websocket/SSE based updating, with possibly a fallback to a single live-data request interface for all updates -> markers, players and additional updates like this daylight thing.
With that we can then have a lot more live-updates a lot faster without spamming too many http-requests ^^'

For this reason I want to finish that websocket/SSE switch first, before adding any more live updates on the webapp.

Additionally there is plans to completely rewrite the webapp etc.. all which is tied tightly together with SOO many ideas i have and want to implement. The issue is that most of those ideas and plans i have not written down anywhere ^^' Which is my fault, but it basically makes it impossible for someone to make a PR for bluemap that doesn't clash with any of my plans :/

PR #585, opened 2024-08-04
bluemap-ru

Svova202:


TBlueF:
Hi @Svova202 :)
Thank you for contributing!

However you edited the `en.conf` which means you changed the default english language to be russian instead of adding a new language. 

Also bluemap already has a russian translation here: https://github.com/BlueMap-Minecraft/BlueMap/blob/master/BlueMapCommon/webapp/public/lang/ru.conf
So you should edit that file instead if you find anything to improve.

*(Edit: If you wanted to just change your own translation on your own bluemap instance, then you don't have to create a Pull Request here for that, you can just edit the files in your `./bluemap/web/lang` folder on your server!)*

Svova202:
Thanks you!

PR #595, opened 2024-08-14
add logs the detailed error message in php

ColdeZhang:
It really takes me time to find out why i can't host web standalone with just "Failed to connect to database".

Then i added the log detail, everything just be cleared: `NOTICE: PHP message: could not find driver`

So I think it's necessary to log err details for solving problems...

### DIFF
--- a/BlueMapCommon/webapp/public/sql.php
+++ b/BlueMapCommon/webapp/public/sql.php
@@ -150,7 +150,8 @@ if (startsWith($path, "/maps/")) {
         $sql = new PDO("$driver:host=$hostname;port=$port;dbname=$database", $username, $password);
         $sql->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
     } catch (PDOException $e ) { 
-        error(500, "Failed to connect to database: ".$e->getMessage());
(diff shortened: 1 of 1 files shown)


TBlueF (approved):

TBlueF (changes requested):
Actually, can we add this to the other errors thrown in this php-file as well, please? :)

ColdeZhang:
> Actually, can we add this to the other errors thrown in this php-file as well, please? :)

That's a reasonable idea. I will made this PR modified soon.

TBlueF (changes requested):
Ah sorry,.. but we can not print the full error in the response that is actually sent back to the client, because the error could contain sensitive information.
What i meant was to still use the `error_log()` method you used in the first commit, just for every error, while still sending the generic error message back to the browser =)

ColdeZhang:
> Ah sorry,.. but we can not print the full error in the response that is actually sent back to the client, because the error could contain sensitive information.
> 
> What i meant was to still use the `error_log()` method you used in the first commit, just for every error, while still sending the generic error message back to the browser =)

Oh, sry I forgot that. Wait a minute

TBlueF (approved):
Thanks a lot! :)

PR #615, opened 2024-10-09
Automatically set translation according to browser settings

NikitaCartes:
Second commit fixes problem with languages not always being saved correctly. Can be recreated in demo.

Added `useBrowserLanguage` setting in `web/lang/settings.json`. If set to false, behavior will be as now.

Language selection algorithm:
1) Take the first element of `navigator.languages`
2) If it appears in list of available languages, set locale to it (this is why few locales have been renamed to comply with RFC 5646)
3) If not, take base language (`en` in `en-US`) and find locale that starts with it
4) If such a locale isn't found, repeat for the rest of `navigator.languages`


NikitaCartes:
Based on [this](https://discord.com/channels/665868367416131594/771451216499965953/1074055018505846896) message

TBlueF (approved):
Looks great! Thanks!

PR #632, opened 2024-12-14
Fix fallback languages not loading (#631)

### LINKED ISSUE #631: Unlocalised strings for non-English languages show up as localisation keys when the browser sends a non-English `Accept-Language` header, affecting both auto-detected and set-via-preference-pane webapp languages
### Steps to reproduce
* Use a browser that sends a non-English `Accept-Language` header. For instance, `ja-JP` or `ja` in my case. ※ You can also set this using Chromium DevTools.
* Open a version of the webapp that contains commit e59d035e7cedff6f414e0d73ee3c883a28530f55. Both `master` (7392447a350813ac80b8dfeb6d374cb1c4781a78) and stable 5.5 are affected, including the [demo instance](https://bluecolored.de/bluemap/) as of this writing. (Also affects my backport which includes the same commit, but that's not really relevant here.)
* Open the side menu and notice that strings that unlocalised strings show up not in English, but in their key form.
* Open the preference pane and change the language to English.
* Now change it back to whatever language you had before. (※ Japanese in my case)
* Previously untranslated keys will now show up correctly as English strings.
* Refresh the webapp.
* Observe that untranslated strings are in their key form again.

### Expected result
* Untranslated strings should appear as English strings, not as localisation keys.

### Actual result
* Localisation keys appear instead of English strings for untranslated strings.

### Context
**BlueMap Version:** 7392447a350813ac80b8dfeb6d374cb1c4781a78
**※ Also affects the demo instance:** https://bluecolored.de/bluemap/

### Screenshots (JP)
image
image
image

### Screenshots (KR)
image
image
image

NikitaCartes:
Fix https://github.com/BlueMap-Minecraft/BlueMap/issues/631

1) If enabled and available, the browser's language will be used
2) Otherwise, the language from `settings.default` will be used
3) If neither of the above is available, English will be used

### DIFF
--- a/common/webapp/src/i18n.js
+++ b/common/webapp/src/i18n.js
@@ -14,20 +14,24 @@ export const i18nModule = createI18n({
 export const i18n = i18nModule.global;
 
 export async function setLanguage(lang) {
+	loadLanguage(lang);
+
+	i18n.locale.value = lang;
+
+	document.querySelector('html').setAttribute('lang', lang);
+
+	return nextTick();
+}
+
+async function loadLanguage(lang) {
 	try {
-		if (!i18n.availableLocales.includes(lang)){
+		if (!i18n.availableLocales.includes(lang)) {
 			let messages = await fetchHocon(`./lang/${lang}.conf`);
 			i18n.setLocaleMessage(lang, messages);
 		}
-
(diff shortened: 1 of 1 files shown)


TBlueF:
Thanks! :)

PR #646, opened 2025-01-20
Add vanilla entity models

MiraculixxT:
Add more entity models to the new entity renderer by BlueMap. Draft until most entities are covered (or i run out of time)

**Entities in this PR**
- Chicken (adult/baby)
- Pig (adult/baby)
- Cod
- Salmon
- Tropical fish
- Pufferfish
- Bee (adult/baby)
- Llama (all variants + chest + carpets)
- Fox (orange, white, missing baby)
- Breeze
- Zombie


TBlueF:
I allowed myself to commit a proposal in here on how to handle the entity-data based models for now:
image
=)

I think i prefer handling it in this (hardcoded) way for now, instead of inventing an entirely new and complex data-driven model.
Mostly because a data-driven approach would:
- still need a ton of hardcoded helpers/value-loaders to load the correct state/data from a specific entity-type
- be overly complex to implement
- most likely overengineered for the actual use it would be (i don't expect many resource-packs actively supporting bluemaps custom entity-model format)

Also if/once mojang decides to add an official data-driven format for entity-models, we will need to switch to that.

I think this "hardcoded" way of adding a simple custom renderer per entity-type is the easiest and most flexible way right now :)

Any thoughts about this are still very welcome!

TBlueF:
I think we can close this for now, in favor of https://github.com/BlueMap-Minecraft/BlueMapEntities :D

PR #674, opened 2025-03-22
Add and fix some translation in Vietnamese

songnguxyz:
Added a few strings and modified some.

### DIFF
--- a/common/webapp/public/lang/vi.conf
+++ b/common/webapp/public/lang/vi.conf
@@ -7,7 +7,7 @@
   map: {
     unloaded: "Không có bản đồ."
     loading: "Đang tải bản đồ..."
-    errored: "Có lồi khi tải bản đồ!"
+    errored: "Có lỗi khi tải bản đồ!"
   }
   maps: {
     title: "Bản đồ"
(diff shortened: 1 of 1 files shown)


common/webapp/public/lang/vi.conf
```diff
@@ -141,30 +145,30 @@
...
 </p>
 <p>
 	<h2>Điều khiển cảm ứng:</h2>
 	<table>
-		<tr><th>di chuyển</th><td>chạm + kéo</td></tr>
-		<tr><th>thu phóng</th><td>chạm 2 ngón + nhón</td></tr>
-		<tr><th>xoay/nghiêng</th><td>chạm 2 ngón + di chuyển / xoay</td></tr>
+		<tr><th>Di chuyển</th><td>chạm + kéo</td></tr>
+		<tr><th>Thu phóng</th><td>chạm 2 ngón + nhón</td></tr>
+		<tr><th>Xoay/nghiêng</th><td>chạm 2 ngón + di chuyển / xoay</td></tr>
```
TBlueF, line 166: Is there a specific reason for this change? 
Those words are lowercase in the reference-language (english) as well :)

songnguxyz: Well yes, but most of the time, it's how we mostly format tables in Vietnamese, as we prefer Uppercase instead. This is not Wikidata actually and we all know it...

Issue #81, opened 2020-09-05
Extend map-limit config capabillities


Blueeyestar:
Since our world is round, we would be happy if BlueMap could also render the world round so it would look more beautiful.

TBlueF:
Meant is a way to limit the map-render bounds in a round shape.

Shanwer:
You can try worldboarder first.

Blueeyestar:
> 
> 
> You can try worldboarder first.

As far as I know, the plugin does not yet support worldborder. Or is there already other information?

TBlueF:
No worldborder support.
I think they meant if you trim your world to be generated round (via worldborder) the map will also be round.

Shanwer:
> > You can try worldboarder first.
> 
> As far as I know, the plugin does not yet support worldborder. Or is there already other information?

Set a round border,then /wb fill the world,it will generatre round chunks.

Shanwer:
> No worldborder support.
> I think they meant if you trim your world to be generated round (via worldborder) the map will also be round.

Yep that's what I meant.

2beFlo:
I'd like to have that feature, too. My main map dates back to 2014, was filled round with wb and over the years got some quite irregular edges. So beeing able to hide them would be awesome :)

TBlueF:
Maybe support for more complex shapes as well?
(Requested on discord)

2beFlo:
Are there any plans left to implement at least the round map feature? It would still be greatly apreciated by me and my server.

TBlueF:
As long as this issue isn't closed, there are plans, yes ... 

I am **roughly** going by the order of the [TODO-List](https://github.com/orgs/BlueMap-Minecraft/projects/2) .. 
progress is being made, but i am just one person ^^

PR #692, opened 2025-05-01
feat: Detect when player is within map bounds

zachthedev:
### Issue
If there are multiple maps, the player shows up on the lowest priority map of the same dimension but not where the rendered part of chunks are.

### Use case
A nether map with roof transparency and then another map that is only the roof and above (to showcase redstone builds, etc).

- Clicking on a player on a different map, will take the user to the correct map for viewing what they are working on

### DIFF
--- a/common/webapp/src/components/Menu/MarkerItem.vue
+++ b/common/webapp/src/components/Menu/MarkerItem.vue
@@ -75,7 +75,7 @@ export default {
 
       if (this.marker.type === "player") {
 
-        if (this.marker.foreign) {
+        if (this.marker.foreign || this.marker.outOfBounds) {
           let matchingMap = await this.$bluemap.findPlayerMap(this.marker.playerUuid);
           if (!matchingMap) return;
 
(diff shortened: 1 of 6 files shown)


TBlueF (changes requested):
I actually forgot to review this PR, sorry ..
Here are my thoughts:

The implementation looks good! There is just the problem that this implementation will clash with the plans to make the map-render-boundaries more versatile: https://github.com/BlueMap-Minecraft/BlueMap/issues/81

The plan here is to make it possible to add all kinds of shapes which can be defined in the map-config for in or excluding them, and even make it possible to extend this with addons.

To make the webapp compatible with this, i would suggest a server-side approach .. where instead of the webapp knowing the map-boundaries and having it calculate if the player is out of bounds, send the player.outOfBounds from the server and let the server calculate that boolean.
This way the server can easily be extended by more complex map-boundaries and the webapp doesn't need to mirror all that logic.

zachthedev:
@TBlueF No worries! That totally makes sense. I had actually started on that approach to refactor this PR, but time got away from me. I plan to pick it back up sometime in the next week or so.

TBlueF:
BlueMap has now implemented `render-masks` which this PR would need to be updated to.
Additionally there should be a configuration to turn this behavior on and off.

I will close this for now due to inactivity, feel free to re-open if this is being updated :)

PR #720, opened 2025-07-13
Atlas support and (resource)pack improvements

TBlueF:


database64128:
Hi, sorry for posting this comment out of nowhere. This is not a bug report, just something to maybe keep in mind?

It'd be nice if the name of the render thread does not change again in the future. I wrote this little script last year to pin BlueMap's render threads to E cores, because I wanted to keep the P cores for other tasks, and it's less noisy (less heat -> lower AIO cooler fan RPMs) to render on E cores. The script was broken after upgrading to the latest version, because this PR changed the name of the render thread.

```sh
#!/usr/bin/env fish

# Get the TIDs of threads with the name "RenderManager-*"
set tids (ps -eLo tid,comm | grep 'RenderManager-' | awk '{print $1}')

# Loop over each TID
for tid in $tids
    # Execute 'taskset -p ffff00 <tid>' on each TID
    taskset -p ffff00 $tid
end
```

_Update: In case it's also useful for someone else, here's the updated script:_

```sh
#!/usr/bin/env fish

set atom_file "/sys/devices/cpu_atom/cpus"

if not test -f $atom_file
    echo "Hybrid P/E cores not detected; nothing to pin."
    exit 0
end

set atom_list (string trim (cat $atom_file))

if test -z "$atom_list"
    echo "/sys/devices/cpu_atom/cpus is empty; nothing to pin."
    exit 0
end

echo "Using E-core CPU list: $atom_list"

# Match RenderManager- (pre-5.10) and BlueMap-RenderT (v5.10 and onwards)
set tids (ps -eLo tid,comm | grep -E 'RenderManager-|BlueMap-RenderT' | awk '{print $1}')

if test -z "$tids"
    echo "No BlueMap render threads found"
    exit 1
end

for tid in $tids
    taskset -pc $atom_list $tid
end
```

TBlueF:
@database64128 Sorry, but I can't make any such guarantees. There is no change like this planned currently but there still might be changes at any point if it is necessary to implement new things, maintain bluemap or just keep things tidy :)
BlueMap-Core and Common can have breaking changes of any nature with any update. Otherwise i'd need to freeze bluemaps development ^^

database64128:
That's quite all right! It's easy for me to notice and fix the script. Should there be another change to the thread name, maybe make sure it's mentioned in the release notes?

TBlueF:
I will try to remember :+1: ,.. but honestly this is a cool, but *very niche* usecase you have there :D
If i'd go into that much detail for every change i do i could just post the git-diff as the changelog ^^

PR #732, opened 2025-08-19
Fix incorrect onPlayerLeave handling in EventForwarder.java inside PaperMC

nicolasvac:
Hi everyone, not really sure if this is a bug or intended, just found it while reading the code.


TBlueF (approved):
Good catch ^^
The only listener using this event is doing the exact same action on join and on leave, which is why this wasn't causing any trouble so far :D

PR #745, opened 2025-10-21
Switch to webserver using a virtual-thread per connection

TBlueF:
This PR reimplements the webserver for a better to read implemementation using blocking-io but a virtual-thread per connection.

This PR should only be merged once BlueMap depends on Java 24 or higher. In earlier Java versions the virtual-threads can lead to deadlocks when accessing a database due to neccesary `synchronized` blocks in the database connection-pool leading to [thread-pinning](https://docs.oracle.com/en/java/javase/21/core/virtual-threads.html#GUID-704A716D-0662-4BC7-8C7F-66EE74B1EDAD).
Thread-Pinning has been fixed in Java 24.


TBlueF:
merged with #779

PR #748, opened 2025-11-04
Update settings.conf

7scarr:


TBlueF:
If you want to change this, please change it in your own bluemap installation. In your server folder look for ./bluemap/web/lang/settings.json. You can change it there :)
This here would change the default language for everyone who is using bluemap, and we will not change that ^^

PR #762, opened 2025-12-24
Allow browsers to cache files

Macil:
This PR replaces cache-busting query parameters on all file requests to allow browsers to cache files. With this PR, if you reopen a BlueMap page that you've opened in the same area before and the assets haven't changed, then barely any network traffic will happen (ie. <1 mb instead of 15+ mb for a short peek), even if you refresh or press the "update map" button.

The browser will automatically use any cached responses that are still fresh ("fresh" meaning that the cached response is younger than its Cache-Control header's max-age value), and will issue conditional requests for any stale cached assets **if** the web server delivered the assets with an ETag or Last-Modified header (as most setups with an external webserver will do, including the setups described by the [External webservers](https://bluemap.bluecolored.de/wiki/webserver/ExternalWebserversFile.html) and [Cloudflare](https://bluemap.bluecolored.de/community/CloudflareR2.html) guides), saving bandwidth when nothing has changed by allowing the server to respond with an empty 304 response.

The "update map" button and requests made after a refresh now work by using Fetch's `cache: "no-cache"` option, which is superior to cache-busting because it allows the browser to make conditional requests if it has any (fresh or stale) cached responses.

If the web server (like BlueMap's built-in webserver) does not include ETag or Last-Modified headers in its responses, then Fetch with the `cache: "no-cache"` option works pretty similar to the current cache-busted queries.

There is one possible visible downside with this PR: on an initial non-refresh page load, the browser may use any fresh cached assets immediately without re-validating against the server, which is great for load times but may be bad if the assets were served with a large Cache-Control max-age value. BlueMap's built-in webserver sets the Cache-Control max-age to 24 hours, so someone re-opening a BlueMap within 24 hours may see assets that are up to 24 hours old (until they refresh or press the "update map" button). If this is an issue, I suggest changing BlueMap's built-in webserver to serve assets with a shorter max-age value, such as 1 hour or 10 minutes.

The one kind of hairy part of this PR is RevalidatingFileLoader.js, which is forked from Three.js's own FileLoader because it has multiple issues preventing it from being used here for this purpose (it doesn't support passing extra options to Fetch, and it can improperly de-dupe requests for the same URL made before and after pressing the "update map" button). The changes to it are pretty light and it seems to only use public APIs of Three.js, so I think it's a decent route. (I do also have an alternative hacky implementation of RevalidatingFileLoader at https://github.com/Macil/BlueMap/blob/betterCaching-monkeyPatch/common/webapp/src/js/util/RevalidatingFileLoader.js which is instead a wrapper around Three.js's FileLoader with some ugly monkey-patching. I don't trust this hacky route to be as future-proof.)

### DIFF
--- a/common/webapp/.eslintrc.cjs
+++ b/common/webapp/.eslintrc.cjs
@@ -1,6 +1,9 @@
 /* eslint-env node */
 module.exports = {
   root: true,
+  env: {
+    es2022: true
+  },
   'extends': [
     'plugin:vue/vue3-essential',
     'eslint:recommended'
--- a/common/webapp/src/js/BlueMapApp.js
+++ b/common/webapp/src/js/BlueMapApp.js
@@ -26,13 +26,14 @@ import "./BlueMap";
 import {MapViewer} from "./MapViewer";
 import {MapControls} from "./controls/map/MapControls";
 import {FreeFlightControls} from "./controls/freeflight/FreeFlightControls";
-import {FileLoader, MathUtils, Vector3} from "three";
+import {MathUtils, Vector3} from "three";
 import {Map as BlueMapMap} from "./map/Map";
-import {alert, animate, EasingFunctions, generateCacheHash} from "./util/Utils";
+import {alert, animate, EasingFunctions} from "./util/Utils";
 import {MainMenu} from "./MainMenu";
 import {PopupMarker} from "./PopupMarker";
 import {MarkerSet} from "./markers/MarkerSet";
 import {getLocalStorage, round, setLocalStorage} from "./Utils";
+import {RevalidatingFileLoader} from "./util/RevalidatingFileLoader";
 import {i18n, setLanguage} from "../i18n";
 import {PlayerMarkerManager} from "./markers/PlayerMarkerManager";
 import {NormalMarkerManager} from "./markers/NormalMarkerManager";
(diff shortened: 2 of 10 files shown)


TBlueF:
I didn't know `no-cache` was a thing for requests, thats useful :)

On a really quick first test it seems to work fine :+1: 
Only thing that is weird to me is that apparently the header is not shown in the browsers dev-tools :thinking: 
image
As i said, it still seems to work and make the browser ignore its cache .. Do you have an idea why its not showing? :thinking:

TBlueF:
> Most HTTP/1.0 caches don't support no-cache directives, so historically max-age=0 was used as a workaround. But only max-age=0 could cause a [stale response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Caching#fresh_and_stale_based_on_age) to be reused when caches disconnected from the origin server. must-revalidate addresses that. That's why the example below is equivalent to no-cache.
>
> Cache-Control: max-age=0, must-revalidate

I guess the browser could be just transforming `no-cache` into `max-age=0` .. and that's why

Macil:
Right, [Fetch's cache option](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#cache) doesn't fully correspond with request header values and `cache: "no-cache"` does lead to `Cache-Control: max-age=0` being in the request headers.

TBlueF:
Thanks for this!

PR #763, opened 2025-12-24
Use raw mouse input in free flight mode

Macil:
This PR uses requestPointerLock's unadjustedMovement option to enable raw mouse input in full-screen free flight mode.

- Chrome supports reading raw mouse input when pointer lock is used, and it seems to be generally recommended for first-person controls to make the mouse movement's relation to your looking angle more predictable: https://web.dev/articles/disable-mouse-acceleration.
- This setting also happens to work around a Chrome bug I encounter very often. When using free flight mode in Chrome, I noticed sometimes while I looked around with the mouse, over a single frame my viewpoint would suddenly rotate far, leaving me disoriented looking in a completely different direction. I debugged this by making `MouseRotateControls.onMouseMove` log the values of `evt.movementX`, and I noticed the issue was caused by `evt.movementX` suddenly having an extreme value like -720 out of nowhere instead of its normal range of values (roughly -40 to 40). I get an incident like this about once a minute when looking around a lot in Chrome, and never in Firefox.

  I found that this is a commonly known issue in Chrome (https://github.com/wasm-bindgen/wasm-bindgen/issues/3862, https://github.com/mrdoob/three.js/issues/12757, https://github.com/Triang3l/WebQuake/issues/37, https://issues.chromium.org/issues/40547981), and this issue can be worked around by using requestPointerLock's unadjustedMovement option. It's convenient that the option that's good for first-person controls also happens to be the one that fixes this problem.

### DIFF
--- a/common/webapp/src/js/controls/freeflight/FreeFlightControls.js
+++ b/common/webapp/src/js/controls/freeflight/FreeFlightControls.js
@@ -152,8 +152,16 @@ export class FreeFlightControls {
 
         document.body.requestFullscreen()
             .finally(() => {
+                // try with unadjustedMovement first and fall back without it if not supported
                 this.target.requestPointerLock({
                     unadjustedMovement: true
+                })
+                .catch(err => {
+                    if (err.name === "NotSupportedError") {
+                        return this.target.requestPointerLock();
+                    } else {
+                        throw err;
+                    }
                 });
             });
     }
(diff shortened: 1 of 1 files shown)


TBlueF:
I am getting (on chrome w/ linux mint) this error with this change:
image
And the pointer-lock is not aquired.

So i am guessing there'd need to be some fallback ..
Also on how many browsers have you tested this? Does it still work fine on all browsers? My worry is that it (when it is supported and doesn't throw an error) could mess with the sensitivity on some systems :)

Macil:
Whoops, I thought the worst case scenario was that a browser that didn't recognize it would ignore it (which was the behavior I saw in Firefox), but apparently browsers that recognize the option may not always support it. Now that I re-read https://web.dev/articles/disable-mouse-acceleration more carefully, I see that they have an example trying again without the option if it failed with it. I've adapted this PR to do that.

PR #767, opened 2026-01-12
Extremely minor spelling fix

TechnicJelle:
:sweat_smile: I just noticed this and couldn't help myself


TBlueF:
I have this weird muscle memory to add the `'` every time I type `its` or similar things, even if i know exactly when it should be used and when it shouldn't .. it's really annoying D:
Thanks for the fix ;D

PR #778, opened 2026-03-11
Add/Improve SQLite compatiblity

Soapy7261:
This PR aims to add or improve compatibility/usage for SQLite databases.
While the BlueMap base already technically supports SQLite, certain parts (such as the sql.php script) do not.

Compatibility todo:
- [x] sql.php
- [x] BlueMap base

Options that should be added, in order of how important they probably are (None of them should require any internal code changes?):
- [x] ~~Option for enabling WAL (`PRAGMA journal_mode = WAL;`, definitely most important imo)~~
- [x] ~~Option(s?) for increasing cache size (`PRAGMA cache_size = -20000;`, negative means kilobytes, so 20MB cache, should be configurable though)~~
- [x] ~~Option for decreasing syncs (`PRAGMA synchronous = NORMAL;`, only safe to use with WAL, might lose the latest transactions on sudden power loss, but shouldn't corrupt the DB itself)~~
- [x] ~~Option for storing temporary tables in RAM (`PRAGMA temp_store = memory;`, not necessary with proper indexing)~~

^ All options can be added by the user now

Other todo:
- [x] Limit max-connections to 1 if using SQLite (Using more doesn't have any performance benefit in my testing, and is just more likely to hit annoying `SQLITE_BUSY` errors, even if using WAL)
- [ ] All `@TODO`s solved

### DIFF
--- a/common/webapp/public/sql.php
+++ b/common/webapp/public/sql.php
@@ -149,7 +149,7 @@ if (startsWith($path, "/maps/")) {
     // Initialize PDO
     try {
         if ($driver === "sqlite") {
-            $sql = new PDO("$driver:file:$sqlfile?mode=ro");
+            $sql = new PDO("$driver:file:$sqlfile");
         } else {
             $sql = new PDO("$driver:host=$hostname;port=$port;dbname=$database", $username, $password);
         }
--- a/common/webapp/public/sql.php
+++ b/common/webapp/public/sql.php
@@ -3,7 +3,7 @@
 // !!! SET YOUR SQL-CONNECTION SETTINGS HERE: !!!
 
 $driver   = 'mysql'; // 'mysql' (MySQL), 'pgsql' (PostgreSQL) or 'sqlite' (SQLite)
-$sqlfile  = 'bluemap.db'; // Only applies when using 'sqlite' as the driver
+$sqlfile  = 'map.db'; // Only applies when using 'sqlite' as the driver
 $hostname = '127.0.0.1';
 $port     = 3306;
 $username = 'root';
--- a/common/webapp/public/sql.php
+++ b/common/webapp/public/sql.php
@@ -2,7 +2,8 @@
 
 // !!! SET YOUR SQL-CONNECTION SETTINGS HERE: !!!
 
-$driver   = 'mysql'; // 'mysql' (MySQL) or 'pgsql' (PostgreSQL)
+$driver   = 'mysql'; // 'mysql' (MySQL), 'pgsql' (PostgreSQL) or 'sqlite' (SQLite)
+$sqlfile  = 'bluemap.db'; // Only applies when using 'sqlite' as the driver
 $hostname = '127.0.0.1';
 $port     = 3306;
 $username = 'root';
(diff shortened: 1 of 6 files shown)


common/webapp/public/sql.php
```diff
@@ -2,7 +2,8 @@
 
 // !!! SET YOUR SQL-CONNECTION SETTINGS HERE: !!!
 
-$driver   = 'mysql'; // 'mysql' (MySQL) or 'pgsql' (PostgreSQL)
+$driver   = 'mysql'; // 'mysql' (MySQL), 'pgsql' (PostgreSQL) or 'sqlite' (SQLite)
+$sqlfile  = 'bluemap.db'; // Only applies when using 'sqlite' as the driver
```
Soapy7261: Comment for future self: This should probably specify it'll (most likely) require an absolute path instead of a relative path

Soapy7261: Should also specify that you'll need an extra driver on Windows for SQLite on PHP

TBlueF:
> Probably wrong repo for this?  

nope, correct repo

My thoughts about the PRAGMA statements:
If these get implemented then i think i would like them to be somehow completely configurable .. as in ..
Some configuration e.g. similar to `connection-properties` that allows the user to specify key and value of each pragma they want to configure.

I also don't like the thought of having sqlite-specific settings in the `sql.conf` that are being ignored by all other database-types ..
Maybe we need to split SQL and Sqlite storage-types and make a separate config-template for sqlite for this.
`max-connections` could then also be hidden and default to 1 in the sqlite config-template.

Soapy7261:
(noted!)
When I last used SQLite, it required a separate driver jar and proper driver-class specification, if we are going to make a separate `sqlite.conf`, maybe a library should be built-in for it to work natively?

TBlueF:
No, bluemap doesn't include any of the sql-drivers. 
If we include one, we need to include all of them,
and i don't think i want to do that ^^

Soapy7261:
Oh, I didn't realize none were bundled.. oops?
I'll add a valid JDBC driver URL for SQLite to the sqlite.conf comments then (similar to how mariadb is already commented)

Soapy7261:
While technically less configurable, I don't think there's really any reason to include the whole `connection-url` as a config option for SQLite, as it's kind of confusing as-is with like `jdbc:sqlite:file:bluemap/bluemap.db`, and any options I've seen come *after* the `jdbc:sqlite:file`, so maybe the option should just be like `filename` and used like `bluemap/bluemap.db` or something? You could still pass flags at the end too.
Perhaps it's better that way?

Soapy7261:
Works for SQLite now and respects all PRAGMAs I tested, solving all `@TODO`s is all that's left (I think?)

Soapy7261:
~~Might also make sense to have a config option to run a `PRAGMA integrity_check` (or maybe `PRAGMA quick_check`?) to make sure the DB isn't corrupt before trying to use it.~~
~~Or just handle `PRAGMA (integrity)/(quick)_check`~~
Or.. don't

TBlueF:
Implemented with 5eef58ace73a615a2ae75fbf30546d95f5a68b02 :)

PR #780, opened 2026-03-18
Linear v2 support

xKumorio:


TBlueF:
Thank you for the PR! 
However, as discussed on Discord, support for the Linear format should be moved out of BlueMap into an addon instead :)

PR #803, opened 2026-05-12
Implement default PlayerDisplayNameProvider API

TechnicJelle:
Implemented a default for the API added in https://github.com/BlueMap-Minecraft/BlueMapAPI/pull/9

I'm not all too happy about it all yet, especially the locations of some things and what I had to pass into other places.
Also, the default implementation currently kind of sucks.

I'd be happy to get feedback on this!

<details>
<summary>Screenshots of this working</summary>

image

image

image

</details>

<details>
<summary>Addon plugin code</summary>

```java
package com.TechnicJelle.blueMapCustomPlayerNames;

import de.bluecolored.bluemap.api.BlueMapAPI;
import net.kyori.adventure.text.Component;
import net.kyori.adventure.text.TextComponent;
import org.bukkit.entity.Player;
import org.bukkit.plugin.java.JavaPlugin;

import java.util.function.Consumer;

public final class BlueMapCustomPlayerNames extends JavaPlugin {

    @Override
    public void onEnable() {
        BlueMapAPI.onEnable(bluemapOnEnable);
    }

    Consumer<BlueMapAPI> bluemapOnEnable = (BlueMapAPI api) -> {
        api.getPlugin().setPlayerNameProvider(playerUUID -> {
            Player player = getServer().getPlayer(playerUUID);
            if (player == null) return playerUUID.toString();

            Component displayName = player.displayName();
            if (displayName instanceof TextComponent textComponent)
                return "<span style=\"color: blue;\">" + textComponent.content() + "</span>";

            return playerUUID.toString();
        });
    };
}
```

</details>


common/src/main/java/de/bluecolored/bluemap/common/plugin/ServerPlayerDisplayNameProvider.java
```diff
@@ -0,0 +1,53 @@
...
+    public ServerPlayerDisplayNameProvider(Server server) {
+        this.server = server;
+    }
+
+    @Override
+    public String get(UUID playerUUID) {
+        // TODO: I am not happy with needing to do this here.
+        //  The place that calls this function already loops over the online players, so this loops double!
+        //  I can't really think of a better way to do this right now, though.
+        for (Player onlinePlayer : this.server.getOnlinePlayers()) {
```
TBlueF: I would be fine with changing the `Server#getOnlinePLayers()` method into a `Map<UUID, Player>` .. this would then also need to be implemented in all implementations though ofc..

common/src/main/java/de/bluecolored/bluemap/common/live/LivePlayersDataSupplier.java
```diff
@@ -41,14 +42,16 @@
...
     private final Server server;
     private final PluginConfig config;
     private final World world;
     private final Predicate<UUID> playerFilter;
 
     private transient @Nullable ServerWorld serverWorld;
 
-    public LivePlayersDataSupplier(Server server, PluginConfig config, World world, Predicate<UUID> playerFilter) {
+    public LivePlayersDataSupplier(Plugin plugin, Server server, PluginConfig config, World world, Predicate<UUID> playerFilter) {
+        this.plugin = plugin;
```
TBlueF: I already don't like passing the `PluginConfig` here, now also passing the `Plugin` screams that this needs to be refactored somehow that this is not needed.
For now, passing the PlayerDisplayNameProvider directly would be the most passable i think ..

common/src/main/java/de/bluecolored/bluemap/common/plugin/ServerPlayerDisplayNameProvider.java
TechnicJelle: Done in 98a5740312a05404ccf44dbf9c049502796eb48e!

common/src/main/java/de/bluecolored/bluemap/common/live/LivePlayersDataSupplier.java
TechnicJelle: Done in 69192028c6b932335dadeab26f70a08e4fb3d283!

common/src/main/java/de/bluecolored/bluemap/common/plugin/Plugin.java
```diff
@@ -206,7 +208,7 @@ private void load(@Nullable ResourcePack preloadedResourcePack) throws IOExcepti
                         MapRequestHandler mapRequestHandler;
                         BmMap map = maps.get(id);
                         if (map != null) {
-                            mapRequestHandler = new MapRequestHandler(map, serverInterface, pluginConfig, Predicate.not(pluginState::isPlayerHidden));
+                            mapRequestHandler = new MapRequestHandler(playerDisplayNameProvider, map, serverInterface, pluginConfig, Predicate.not(pluginState::isPlayerHidden));
```
TBlueF: Yeah .. so this wont work either, since the variable is not initialized here, and if you change the variable via API after bluemap is loaded, it won't update either.. 😅
I'll take a look at this in a bit, rn i don't like any of the "easy" solutions to pass the information through to the LivePlayerDataProvider correctly..

TechnicJelle:
Yep, still works: 
image
image

PR #806, opened 2026-05-31
fix: replace non-breaking space with regular space - in Unicode, it is represented as U+00A0, whereas a standard space bar hit is U+0020

ashwalk33r:
The character I replaced is a Non-Breaking Space (often abbreviated as NBSP).
In Unicode, it is represented as U+00A0, whereas a standard space bar hit is U+0020.
To the naked eye, it looks exactly like a regular space, but it behaves very differently under the hood.

A standard space tells a browser or word processor, "You can break the line here if you run out of room." A Non-Breaking Space does the exact opposite. It creates a space between two words but forces them to stay glued together on the same line, preventing an awkward line break (word wrap).

It is redundant and is problematic to work with thus I've replace it repo-wide with a regular space.

---

I've found it by browsing lang files as VS Code highlighted it in distinct way (look at the "yellow orange-ish" in the line 53):

image

### DIFF
--- a/common/webapp/public/lang/id.conf
+++ b/common/webapp/public/lang/id.conf
@@ -1,175 +1,175 @@
 {
-  pageTitle: "BlueMap - {map}"
-  menu: {
-    title: "Menu"
-    tooltip: "Menu"
-  }
-  map: {
-    unloaded: "Peta tidak tersedia."
-    loading: "Memuat peta..."
-    errored: "Terjadi kesalahan saat memuat peta ini!"
-  }
-  maps: {
-    title: "Peta-Peta"
-    button: "Peta-Peta"
-    tooltip: "Daftar Peta"
-  }
-  markers: {
-    title: "Penanda"
-    button: "Penanda"
-    tooltip: "Daftar Penanda"
-    marker: "penanda | penanda"
-    markerSet: "set-penanda | set-penanda"
-    searchPlaceholder: "Cari..."
-    followPlayerTitle: "Ikuti Pemain"
-    sort {
-      title: "Urutkan berdasarkan"
-      by {
-        default: "bawaan"
-        label: "nama"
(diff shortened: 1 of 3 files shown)


TBlueF:
One question before i merge:
I agree that we should use normal spaces for indentations. Inside the translations however - as you said - it does make a difference. Have you confirmed that this doesn't change the text appearance on the browser in a non-favourable way? 
I don't speak polish, so i can't check. I just want to confirm that this wasn't a deliberate choice in the original polish translation :)

ashwalk33r:
Let me know is this helpful - I can add in separate MR automated check + fix commands to be run locally and as CICD quality gates.

TBlueF:
Please don't submit any automated / AI PRs to this repo. Thank you :)

ashwalk33r:
> One question before i merge: I agree that we should use normal spaces for indentations. Inside the translations however - as you said - it does make a difference. Have you confirmed that this doesn't change the text appearance on the browser in a non-favourable way? I don't speak polish, so i can't check. I just want to confirm that this wasn't a deliberate choice in the original polish translation :)

I did not test. Let me know how to enable and see such views and I will test (as by chance I know Polish). What's the minspec (browser versions supported, viewpors and devices supported)? I will test against that.

From language point of view keeping those as NBSP or not makes little if no difference.

ashwalk33r:
> Please don't submit any automated / AI PRs to this repo. Thank you :)

I did it myself. You don't want to remove NBSP? You think the screenshot was done by AI?

TBlueF:
@ashwalk33r All good, no hard feelings please :)
But generally low effort PR's that only change formatting and don't actually fix a bug or add any feature are not desired. It takes longer and more effort to review them than to actually do the changes myself. I hope you understand :+1:

ashwalk33r:
whatever

PR #807, opened 2026-05-31
fix: normalize repo-wide file line endings to LF

ashwalk33r:
## Summary

`core/src/main/resourceExtensions/assets/minecraft/blockstates/conduit.json` was committed with CRLF (Windows) line endings. This normalizes it to LF, matching the rest of the resource files.

All other files in the repo are fine, only that one had the issue.

Done with dos2unix https://linux.die.net/man/1/dos2unix

## Why

The file's content is unchanged — only the line endings differ. Because the committed blob used CRLF, any edit on a Unix/WSL system rewrote it to LF, generating a confusing diff that shows all 6 lines removed and re-added with byte-identical text. Normalizing the endings removes this noise and prevents future churn between contributors on different platforms.

## Changes

- Convert `conduit.json` from CRLF to LF line endings (no content change).

## Verification

- `git diff --ignore-all-space` shows no differences (content is unchanged).
- `grep -c $'\r'` on the file returns 0 (no carriage returns remain).

## Notes

Other files in the repo may have the same CRLF/LF inconsistency in the future. A follow-up could add a `.gitattributes` rule (e.g. `*.json text eol=lf`) to enforce LF and prevent this class of diff going forward.


TBlueF:
Please don't submit any automated / AI PRs to this repo. Thank you :)

ashwalk33r:
> Please don't submit any automated / AI PRs to this repo. Thank you :)

yo bro wtf

Issue #293, opened 2022-02-21
Make compressed site friendly for hosted static sites.


yikerman:
**Is your feature request related to a problem? Please describe.**
One of the benefits of static sites is it can be served through static site hosting providers. However when it comes to gzip compressed sites the wiki redirects the `.json` to `.json.gz` with a `Content-Encoding:gzip` http header. However not every (little) hosting providers have this kind of functionality.

**Describe the solution you'd like**
Instead of the full server side method, the client can have an idea about whether the site is compressed and fetch the gzips automatically.

TBlueF:
This is *(to my knowledge)* not feasible:
If the server doesn't send the `Content-Encoding: gzip` header, then the browser doesn't do the decompressing natively. Which would mean that the web-app needs to manually decompress the tiles through JavaScript. Which *(i tried, and)* is **incredible** slow.

yikerman:
What about [pako](https://github.com/nodeca/pako)? The benchmark indicates it’s only about 3 times slower than zlib itself.

TBlueF:
I am not sure what i tried before, and if i tired it with pako .. i can put this somewhere in the TODO, but testing this will be very low prio. And will probably not happen in the near future (unless someone else PRs it)

yikerman:
I totally understand this. I’ve found an alt btw. Thanks for all the support.

ZacharyDuBois:
I feel this would be extremely nice to have. For example, S3 or B2 storage fronted with a CDN provider such as CloudFront, CDN77, or Bunny.net.

But that also begs the other request, push direct to an external storage and don't maintain a local copy.

fechan:
With the [Compression Streams API](https://developer.mozilla.org/en-US/docs/Web/API/Compression_Streams_API) (which has baseline/widely available support according to MDN), it should be possible to use the browser's native gzip decompression. If support for browsers older than 2023 is desired, then I hear wasm gzip libraries are faster than any pure JS implementation.

In any case, I would find this feature very useful, and I might be interested in implementing it. I'm not too familiar with the BlueMap project structure, but I assume the implementation steps would be:
- In [TileLoader.js](https://github.com/BlueMap-Minecraft/BlueMap/blob/master/common/webapp/src/js/map/TileLoader.js) Detect if the server only sends compressed PRBMs (somehow?)
- Instead of the Three.js FileLoader, fetch the `*.prbm.gz`  ourselves and decompress
[...]

PR #815, opened 2026-07-09
Make it possible to host BlueMap on static web hosts with client decompression mode

fechan:
## What it does
This PR is intended to address #293 by allowing you to host the `web` directory generated by BlueMap on a static web host, such as Github Pages. This is achieved by modifying `web/settings.json` and adding `"clientDecompression": true`.

A Github Pages demo is currently available from the following links (though I may move it soon):
- Demo: https://fredchan.org/shack-village-maps
- GH repo: https://github.com/fechan/shack-village-maps

## Implementation details
There are two places that request files that are stored gzipped at rest: `textures.json` and PRBMs. When `clientDecompression` is on, it will attempt to fetch `<filename>.gz` instead of the normal filename, using `RevalidatingFileLoader`. After setting`RevalidatingFileLoader.setClientDecompression(true)`, the file loader will decompress the gzipped data using the browser-native [Compression Streams API](https://developer.mozilla.org/en-US/docs/Web/API/Compression_Streams_API).

## Possible issues
* Client decompression will not work in the Vite preview environment (i.e. when you use `npm run dev`), since the decompression code errors out with `Failed to read data from the ReadableStream: “TypeError: The input data is corrupted: incorrect header check”`. This seems to be the same issue faced by these folks: https://github.com/monochrome-music/monochrome/issues/320. However, this does not affect the production environment.

### DIFF
--- a/common/webapp/src/js/BlueMapApp.js
+++ b/common/webapp/src/js/BlueMapApp.js
@@ -76,6 +76,7 @@ export class BlueMapApp {
          *      maps: string[],
          *      scripts: string[],
          *      styles: string[]
+         *      clientDecompression: boolean,
          *  }}
          **/
         this.settings = null;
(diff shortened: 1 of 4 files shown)


TBlueF (approved):
I am currently in the process of rewriting the webapp, this feature is already implemented there.
But until that rewrite is done, i think it doesn't do any harm to merge this :)

PR #819, opened 2026-07-18
Use Server-sent events to add live updates for files, players, and markers

pR0Ps:
SSE Implementation:
 - A `SseConnectionManager` is created per-map in the `MapRequestHandler`.
 - Whenever a client connects to the `live/sse` endpoint, a new `SseConnection` is created for it and is added to the `SseConnectionManager`.
 - The `SseConnectionManager` receives requests to push updates and sends them to all known `SseConnections`. If sending fails, the connection is cleaned up and removed.
 - Sending happens on a background thread so normal server operations are not blocked on clients picking up events.

Backend events:
 - Tile updates are event-based: when a new tile is rendered an event is emitted which causes the tile's X+Y and lod to be broadcasted to all connected clients.
 - Players and other markers are still polling-based, but the polling is done server-side (and only once per endpoint) and connected clients are notified only when the data changes.
 - Polling was implemented by replacing the `CachedRateLimitDataSupplier` with a `LiveDataSupplierBroadcaster`. This `Supplier` wrapper can be used in the same way where requests to `get()` it's data are rate-limited and cached. However, it also runs a background thread that (when listeners are added) will automatically poll the supplier and send events when the data changes.

Frontend changes:
 - On page load, an `EventSource` connects to the `<map>/live/sse` endpoint and watches for events.
   - "tile" updates cause the specified tiles to be force-reloaded
   - "player" updates are fed directly into the existing `playerMarkerManager`
   - "marker" updates are fed directly into the existing `markerFileManager`
 - Since this takes the place of the marker managers polling, the marker managers were updated to be able to be paused, and are started in a paused state.
 - In order to ensure that markers are still updated if the SSE endpoint can't be contacted, the existing polling marker managers will be unpaused if the SSE connection fails to connect or disconnects. If the SSE connection returns, the marker managers will be paused again.
 - Tile force-reloading is handled by removing the specified tile from the `revalidatedUrls` cache, the calling `load` on it. This forces the normal tile loading mechanism to update the tile, reloading it.


common/src/main/java/de/bluecolored/bluemap/common/web/LiveDataSupplierBroadcaster.java
```diff
@@ -0,0 +1,129 @@
...
+    private final Set<Consumer<T>> listeners = ConcurrentHashMap.newKeySet();
+    private final Thread pollThread;
+    private volatile boolean closed = false;
+    private long lastUpdate = -1;
+    private T data = null;
+
+    public LiveDataSupplierBroadcaster(Supplier<T> dataSupplier, long pollIntervalMillis) {
+        this.dataSupplier = dataSupplier;
+        this.pollIntervalMillis = pollIntervalMillis;
+        this.pollThread = new Thread(this::pollLoop, String.format("%sPoller", dataSupplier.getClass().getName()));
```
TBlueF: This will create 2 Platform-Threads for each map that is loaded. Some modded servers have a lot of dimensions and thus a lot of maps, which results in a ton of threads.
Since Platform-Threads are not free, I would love to avoid that somehow.
Maybe we can make use of the `BlueMap.SCHEDULER` here instead?

common/src/main/java/de/bluecolored/bluemap/common/web/SseConnectionManager.java
```diff
@@ -0,0 +1,166 @@
...
+            }
+        }
+        if (!toRemove.isEmpty()) {
+            boolean fire;
+            synchronized (this) {
+                fire = connections.removeAll(toRemove) && connections.isEmpty();
+            }
+            if (fire) notifyHasConnections(false);
+        }
+    }
```
TBlueF, line 103: One doubt i have here is that if one connection is not responsive or very slow, this would probably slow down the update-process for all connections?
Also i would like to have some safe-guard against the `broadcastQueue` growing indefinitely if somehow more events come in than can can be processed/sent :)

Does anything speak against creating a virtual-thread per SseConnection to send the events?
And then maybe each connection could have their own broadcastQueue that can be size-limited.
(Just thinking out loud right now)

TBlueF (changes requested):

common/src/main/java/de/bluecolored/bluemap/common/web/LiveDataSupplierBroadcaster.java
pR0Ps: Good call, I updated this file to just use `BlueMap.SCHEDULER.scheduleWithFixedDelay(...)` to fire the updates, and wrapped the actual call it fires with `BlueMap.THREAD_POOL.execute(...)` so that a slow data supplier doesn't block anything else the scheduler is working on.

common/src/main/java/de/bluecolored/bluemap/common/web/SseConnectionManager.java
pR0Ps: Hmm, yes, a slow client would currently slow down delivery to other clients: `broadcastSync` sends the events to each connection via`SseConnection.send()` which blocks once its internal pipe's buffer is filled up. Unblocking that requires the `HttpResponseOutputStream` to pull the data out of it and write it to the connected client. If the client reads extremely slowly, both the output buffer, as well as the `SseConnection`'s pipe's buffer would fill up, blocking `SseConnection.send()`, blocking `broadcastSync` from moving on to other connections.

I think you're right about the fix:
Since each `SseConnection` is already running in its own virtual thread (they're spawned from `HttpServer.handleConnection()`), I should be able to move the queue + send logic into the `SseConnection` itself. That way the `SseConnectionManager` can broadcast without blocking by just firing the events into each `SseConnection`'s queue and letting it handle draining them to the client. That would also make it easy to implement a per-connection max queue size.

pR0Ps: I've addressed these issues in [6253d80](https://github.com/BlueMap-Minecraft/BlueMap/pull/819/commits/6253d809cccf8f6def1e00d79d3875b1730b4f29), mostly in the way I stated above.

TBlueF (approved):

pR0Ps:
Doing the latest SSE connection refactor made me realize that the way that the SSE data is written could probably be optimized, though it would take some changing of code more on the HTTP sever side of things.

Currently the `SseConnection` uses a virtual thread to push an event into a buffer that the `HttpResponseOutputStream.write` reads out into another buffer, that it then writes to the connected client (via a `BufferedOutputStream`).

This first event buffer was required before the refactor in order to ensure that we could write an event and move on to the next connection without waiting for the client to consume the data. However, since the `SseConnection` is now using a blockable virtual thread to do the writing, there's no reason it needs to write to a buffer at all. We could skip all the buffering entirely if the `SseConnection` could be handed the output stream (after some headers setup) and just block on writing the data directly to that, bypassing everything else.

I *think* it would save the creation of a virtual thread, 2 buffers, some cross-thread overhead, and would avoid having to make [this wider change](https://github.com/BlueMap-Minecraft/BlueMap/pull/819/changes#diff-9421be0abc23f084694af55c9f57a8dc476ec6349b565dd35776a13dcc7da3aaR71). However, it would involve some amount of changes more on the HTTP server side which might be a bit out of scope for at least this initial version.

common/src/main/java/de/bluecolored/bluemap/common/web/SseConnection.java
```diff
@@ -93,14 +93,13 @@ public synchronized boolean setOnClose(Runnable onClose) {
...
     /**
      * Queues an SSE event to be delivered to this connection.
-     *
+     * <p>
      * If this connection's queue is full (due to a slowly-reading client), the event is
      * silently dropped.
      */
-    public void enqueue(String eventType, String data) {
-        if (closed) return;
-        // TODO: close the connection if the queue is full to force a reconnect?
```
pR0Ps: I was thinking it might be a good idea to force a reconnect because if the client has a full queue:
 - If they're actually slowly reading data, the data at the front of the queue is now stale and they might actually benefit from starting again with more up-to-date data.
 - It seems likely that the client is completely stuck/gone and not consuming anything, in which case we should clean up the connection.

Will defer to your opinion on this though.

common/src/main/java/de/bluecolored/bluemap/common/web/SseConnectionManager.java
```diff
@@ -64,16 +69,13 @@ public void add(SseConnection connection) {
...
-            if (connections.add(connection)){
+            if (connections.add(connection)) {
                 if (!connection.setOnClose(() -> remove(connection))){
                     // failed to register an onClose callback (connection is already closed?)
                     connections.remove(connection);
                     fire = false;
                 }
             }
-            else {
-                fire = false;
```
pR0Ps: I had this here as a safety check, but yeah, within this `synchronized` block with an unbounded set I don't see how the `connections.add` can realistically fail.

With this removed, should this now be just:
```java
connections.add(connection);
if (!connection.setOnClose(() -> remove(connection))){
    // failed to register an onClose callback (connection is already closed?)
    connections.remove(connection);
    fire = false;
}
```

TBlueF:
> Doing the latest SSE connection refactor...

@pR0Ps I already had the exact same thought!
So I am actually already looking into if it is feasible to implement what you just suggested :)

Another idea i had was moving the tile-change listeners from the HiresModelManager and the LowresTileManager, over to the Storage implementation. Which would imho be cleaner and a lot more powerful, since Addons then could use Listeners on the GridStorage/ItemStorage as well, and i am not yet 100% happy with having these listeners inside HiresModelManager/LowresTileManager..
Problem with that is that it requires a rather large change in the Storage implementations which would probably also break some addons. So i am still on the fence there..

common/src/main/java/de/bluecolored/bluemap/common/web/SseConnection.java
TBlueF: I wasn't sure about this, since if a client is just very slow, cancelling the connection each time it tries to read something probably won't improve the situation much :D
However i believe you are right that old and incomplete events also won't be very useful, and closing the connection might give the client the chance to recover.
I'd say let's try with going with your suggestion and closing the connection in that case. :+1: 
If it causes problems, people will report them and then we can still change it :)

common/src/main/java/de/bluecolored/bluemap/common/web/SseConnectionManager.java
TBlueF: The only reason i removed this, was because the IDE righfully said that if `connections.add(connection)` returns `false` (-> the `Set` already contains the same connection instance) that means that it can not have been empty here `fire = connections.isEmpty();` .. which means `fire` is guaranteed to already be `false` in that `else` code-path ^^

common/src/main/java/de/bluecolored/bluemap/common/web/SseConnection.java
pR0Ps: The buffer + queue depth can also be tuned so that they're large enough to handle a burst of updates on slow connections. Also, accidently being too eager to disconnect clients wouldn't be that bad - they'll just fall back to using polling for markers and miss some tile updates.

On a related note, an optimization that I was messing with but didn't end up implementing due to the complexity it would've added to this PR was to have the queue be keyed on the event type. That way a player/marker update being pushed could replace any existing update of the same type already in the queue instead of lining up behind it. This would help reduce the queue depth, as well as reduce the latency of useful updates on even slower connections.

I do think it's a worthwhile optimization, but maybe as a follow-up PR once the basic version here is solid.

TBlueF (approved):
Alright, i think i would be fine with merging it like this for now, and then any improvements can be made in follow-up PR's :)

If you agree with that, and with the latest changes i made, then i would press that merge button 👍

pR0Ps:
Agreed, I think this is good to go as-is. All your changes look good, thanks for adding SSE support to the CLI version!

PR #829, opened 2026-08-07
Implement keepalive messages for SSE connections

pR0Ps:
Implements sending an SSE comment (`:`) every 30 seconds if no other event has been enqueued to prevent the connection from timing out and being closed.

### DIFF
--- a/common/src/main/java/de/bluecolored/bluemap/common/web/SseConnection.java
+++ b/common/src/main/java/de/bluecolored/bluemap/common/web/SseConnection.java
@@ -30,6 +30,7 @@ import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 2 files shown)


common/src/main/java/de/bluecolored/bluemap/common/web/SseConnection.java
```diff
@@ -86,16 +92,20 @@ public void enqueue(String eventType, String data) {
...
+        SseEvent event;
         try {
             while (!closed) {
                 try {
-                    event = queue.take();
+                    if (KEEPALIVE_INTERVAL_SECONDS > 0){
+                        event = queue.poll(KEEPALIVE_INTERVAL_SECONDS, TimeUnit.SECONDS);
+                    } else {
+                        event = queue.take();
+                    }
```
pR0Ps, line 103: I kept this `if` statement here even though it's dead code because I figured that if someone attempted to disable the keepalive by setting `KEEPALIVE_INTERVAL_SECONDS = 0` then it should do the right thing vs. continually sending keepalives.

I can remove this and document the constant if that's preferred though.

TBlueF (approved):

PR #862, opened 2026-09-13
Allow adding additional headers for Map/File resources

iCiaran:
Allow adding additional headers by config as mentioned in https://github.com/BlueMap-Minecraft/BlueMap/pull/833#issuecomment-5354302321

Some headers (ETag, Content-Type etc) which are calculated dynamically would break if overridden and cannot be modified.


common/src/main/resources/de/bluecolored/bluemap/config/webserver.conf
```diff
@@ -47,3 +47,12 @@ log: {
   # Default is "%1$s \"%3$s %4$s %5$s\" %6$s %7$s"                         | 10.10.10.10 "GET /assets/file.png HTTP/1.1" 200 OK
   format: "%1$s \"%3$s %4$s %5$s\" %6$s %7$s"
 }
+
+# Additional headers to include in the responses for static files
```
nightowl286: Missing `.` at the end of the comment, not consistent with the other comment lines.

common/src/main/java/de/bluecolored/bluemap/common/plugin/Plugin.java
TBlueF, line 246: I think instead of adding the headers separately to each request-handler, maybe we can move the `BlueMapResponseModifier` from here up to the root (similar to how it's done here: https://github.com/BlueMap-Minecraft/BlueMap/pull/833/changes/c0999f788e6c0b3d50ed39a812d8f6a206942b6d) and then we have to only pass it only to that one class which then takes care of adding them universally..

I would also be in favour of just removing both
```
response.addHeader("Cache-Control", "public");
response.addHeader("Cache-Control", "max-age=" + TimeUnit.DAYS.toSeconds(1));
```
completely in favour of adding them to the default config.

Or am i missing any issues that come with that?

common/src/main/java/de/bluecolored/bluemap/common/web/MapRequestHandler.java
```diff
@@ -28,6 +28,7 @@
 import de.bluecolored.bluemap.common.web.http.HttpStatusCode;
 import de.bluecolored.bluemap.core.map.BmMap;
 import de.bluecolored.bluemap.core.storage.MapStorage;
+import java.util.Map;
```
TBlueF:
```suggestion
```

implementations/cli/src/main/java/de/bluecolored/bluemap/cli/BlueMapCLI.java
```diff
@@ -297,7 +297,7 @@ public void startWebserver(BlueMapService blueMap, boolean verbose) throws IOExc
             routingRequestHandler.register(
                     "maps/" + Pattern.quote(mapConfigEntry.getKey()) + "/(.*)",
                     "$1",
-                    mapRequestHandler
+                    new BlueMapResponseModifier(mapRequestHandler, config.getAdditionalHeaders())
```
TBlueF: not needed, already added on L314

implementations/cli/src/main/java/de/bluecolored/bluemap/cli/BlueMapCLI.java
```diff
@@ -282,7 +282,7 @@ public void startWebserver(BlueMapService blueMap, boolean verbose) throws IOExc
         RoutingRequestHandler routingRequestHandler = new RoutingRequestHandler();
 
         // default route
-        routingRequestHandler.register(".*", new FileRequestHandler(config.getWebroot()));
+        routingRequestHandler.register(".*", new BlueMapResponseModifier(new FileRequestHandler(config.getWebroot()), config.getAdditionalHeaders()));
```
TBlueF: not needed, already added on L314

TBlueF (approved):
LGTM, thanks!

